PR #6, opened 2017-01-10
Fix menu path documentation

focusaurus:
looks like copy-pasta from standard-format


focusaurus:
Someone beat me to this it seems.

PR #14, opened 2017-01-11
converted some to es6

Rafi993:
I did convert some code to es6 but I think it can be done in  more elegent way.

### DIFF
--- a/lib/prettier.js
+++ b/lib/prettier.js
@@ -1,8 +1,8 @@
 /* global atom */
 
-var path = require("path");
-var findRoot = require("find-root");
-var prettier = require("prettier");
+const path = require("path"),
+ findRoot = require("find-root"),
+ prettier = require("prettier");
 
 module.exports = {
   style: null,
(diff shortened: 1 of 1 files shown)


jlongster:
Are all recent Atom version guaranteed to have support for this?

Rafi993:
yes they do. Even the official tutorial https://github.com/blog/2231-building-your-first-atom-plugin is written in es6

lib/prettier.js
```diff
@@ -1,8 +1,8 @@
 /* global atom */
 
-var path = require("path");
-var findRoot = require("find-root");
-var prettier = require("prettier");
+const path = require("path"),
+ findRoot = require("find-root"),
+ prettier = require("prettier");
```
jlongster: Can you do a `const` per line? That's why I prefer.

lib/prettier.js
```diff
@@ -15,12 +15,12 @@ module.exports = {
     "trailingComma",
     "bracketSpacing"
   ],
-  fileSupported: function(file) {
+  fileSupported: (file)=> {
     // Ensure file is a supported file type.
     var ext = path.extname(file);
```
robwise: should this be a `const`?

lib/prettier.js
```diff
@@ -33,35 +33,35 @@ module.exports = {
       this.handleEvents.bind(this)
```
robwise: I think we can remove this line because `handleEvents` is now bound via fat arrow on line 77

jlongster:
We might should go ahead and merge this and do a follow-up. It's going to conflict easily with other ones.

robwise:
Closing this as it is superseded by #22 (which your work inspired 🥇)

PR #15, opened 2017-01-12
added options for arrowParensAlways

oviava:
Added options for `arrowParensAlways` for `ArrowFunctionExpression`


jlongster:
This is going to removed from prettier because that option was never meant to be there in the first place. It's leftover from recast, and is not a large enough stylistic difference for us to have a whole option for it.

PR #16, opened 2017-01-13
Update options in README to reflect current menu

robwise:
The descriptions are from the prettier README. We could, alternatively, just tell people to reference that instead of merging this?


robwise:
Closing, see https://github.com/jlongster/prettier-atom/pull/17 instead

Issue #18, opened 2017-01-13
Allow users to specify extensions to auto format


erbridge:
For instance, we use the `.jsx` extension for JSX in the main project I'm working on, and prettier's JSX support still has some issues. I would like to be able to auto format on save, but that messes up my JSX files.

PR #24, opened 2017-01-15
Remove unsupported file type warning

ryancole:
The unsupported file type warning doesn't seem necessary, especially if format-on-save is enabled. It's understood that prettier formats JS code, so seeing the unsupported file type warning while editing non-JS code isn't that helpful. It gets a bit spammy.

### DIFF
--- a/lib/helpers.js
+++ b/lib/helpers.js
@@ -41,7 +41,6 @@ const format = (event, options = { ignoreSelection: false }) => {
   const isTransformingFile = options.ignoreSelection || !selectedText;
 
   if (isTransformingFile && !isFileTypeSupported(editor.getPath())) {
-    atom.notifications.addWarning('prettier-atom: File type not supported');
     return;
   }
 


robwise:
@ryancole Do you think it should just try and transform it anyway? I worry about having it silently abort like this. Maybe we should remove that whole conditional?

ryancole:
If the file type is not supported, which it looks like is a known conditional via `isFileTypeSupported`, then I'd think you'd just want to abort. You don't want to try to format anyway, because this plugin could be installed on somebody's atom and that person may frequently code in more than just JS. Wouldn't want to run prettier on unsupported formats knowingly.

robwise:
@ryancole Good points! 

> That is, unless prettier itself has a built in knowledge of what it supports

Alas, it does not because we are just feeding it basically a giant string. I just tested it without the guard clause, and it will throw an ugly syntax error without it, so until #18, your way is probably the best way to go, so I'm merging.

PR #32, opened 2017-01-20
tabLength config Atom

oviava:
Original issue was only for tabLength, thought since the option is there as well - include preferredLineLength.

Not sure if this is the best option though but created a map for `prettier` options to atom config key values.


oviava:
think there might be more conflicts in here

lib/helpers.js
```diff
@@ -9,9 +9,28 @@ const PRETTIER_CORE_OPTIONS = [
...
   'bracketSpacing',
 ];
+const CORE_OPTIONS_ATOM_MAP = {
+  printWidth: 'editor.preferredLineLength',
+  tabWidth: 'editor.tabLength',
+};
 
 // helpers
-const getConfigOption = key => atom.config.get(`prettier-atom.${key}`);
+const isFileTypeSupported = path => SUPPORTED_FILE_TYPES.some(ext => path.endsWith(ext));
```
robwise: I think this is the only line that got messed up by my other PR. I switched to using scopes so we don't need this anymore

package.json
```diff
@@ -59,17 +59,47 @@
...
-      "type": "integer",
-      "default": 2,
+      "title": "Tab Width Settings",
+      "description": "Tab Width Settings",
+      "type": "object",
+      "properties": {
+        "useDefault": {
+          "title": "Use Atom Default",
+          "type": "boolean",
+          "description": "Use Atom default default tab length setting",
```
robwise: `Use Atom's default tab length setting`

package.json
```diff
@@ -59,17 +59,47 @@
...
+      },
       "order": 2
     },
     "tabWidth": {
-      "title": "Tab Width",
-      "description": "Number of spaces it should use per tab",
-      "type": "integer",
-      "default": 2,
+      "title": "Tab Width Settings",
+      "description": "Tab Width Settings",
```
robwise: Atom doesn't seem to use the description for objects so we should probably get rid of this line for clarity

package.json
```diff
@@ -59,17 +59,47 @@
...
-      "type": "integer",
-      "default": 80,
+      "type": "object",
+      "title": "Print Width Settings",
+      "description": "Code fit settings",
+      "properties": {
+        "useDefault": {
+          "title": "Use Atom Default",
+          "type": "boolean",
+          "description": "Use Atom default default wrap setting",
```
robwise: `"Use Atom's preferred line length setting"`

package.json
```diff
@@ -59,17 +59,47 @@
       "order": 1
     },
     "printWidth": {
-      "title": "Print Width",
-      "description": "Fit code within this line limit",
-      "type": "integer",
-      "default": 80,
+      "type": "object",
+      "title": "Print Width Settings",
+      "description": "Code fit settings",
```
robwise: Atom doesn't seem to use the description for objects so we should probably get rid of this line for clarity

robwise:
superceded by #55

PR #42, opened 2017-02-06
Update to 0.16.0

arnarthor:
😄


robwise:
Thanks, I will merge. FYI I need to revert the change to `version` because that needs to be done by atom when I publish

arnarthor:
Alright, awesome. I'll keep that in mind next time 👍

PR #44, opened 2017-02-08
add ability to transform js embedded in .html or .vue files

transcranial:
For JS embedded within a `<script>` tag in .html or .vue files, selecting the text and transforming currently works well, but it would be nice if this were easier. I've added `text.html.basic` and `text.html.vue` to the allowed scopes, and for these scopes, added the ability to transform the text only within the `<script>` tag.

### DIFF
--- a/lib/helpers.js
+++ b/lib/helpers.js
@@ -10,6 +10,11 @@ const PRETTIER_CORE_OPTIONS = [
   'bracketSpacing',
 ];
 
+const EMBEDDED_SCOPE = [
+  'text.html.vue',
+  'text.html.basic',
+];
+
 // helpers
 const getConfigOption = key => atom.config.get(`prettier-atom.${key}`);
 
@@ -31,6 +36,17 @@ const executePrettier = (text, formatOptions) => {
   }
 };
 
+const transformRange = (editor, range) => {
+  const cursorPositionPriorToFormat = editor.getCursorScreenPosition();
+  const textToTransform = editor.getTextInBufferRange(range);
+  const transformed = executePrettier(textToTransform, getFormatOptions());
+  if (!transformed) {
+    return;
+  }
+  editor.setTextInBufferRange(range, transformed);
+  editor.setCursorScreenPosition(cursorPositionPriorToFormat);
+};
+
 // public API
 const format = (event, options = { ignoreSelection: false }) => {
   const editor = atom.workspace.getActiveTextEditor();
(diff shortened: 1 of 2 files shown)


lib/helpers.js
```diff
@@ -42,20 +42,48 @@ const format = (event, options = { ignoreSelection: false }) => {
...
   const isTransformingFile = options.ignoreSelection || !selectedText;
 
-  const textToTransform = isTransformingFile ? editor.getText() : selectedText;
+  const currentScope = getCurrentScope();
+  const isEmbedded = ['text.html.vue', 'text.html.basic'].includes(currentScope);
+  if (isEmbedded) {
+    const re = /<script\b[^>]*>([\s\S]*?)(?=<\/script>)/gi
+    editor.scan(re, iter => {
+      // only transform first match
+      iter.stop();
```
robwise: does this mean if there are two script blocks in the file, it only transforms the first one?

lib/helpers.js
```diff
@@ -42,20 +42,48 @@ const format = (event, options = { ignoreSelection: false }) => {
   const selectedText = editor.getSelectedText();
   const isTransformingFile = options.ignoreSelection || !selectedText;
 
-  const textToTransform = isTransformingFile ? editor.getText() : selectedText;
+  const currentScope = getCurrentScope();
+  const isEmbedded = ['text.html.vue', 'text.html.basic'].includes(currentScope);
```
robwise: Is this going to affect how it works when doing a manual format (as opposed to format on save)?

Also, can we put the list of embedded scopes in a constant up at the top of the file so it's easy to find later?

lib/helpers.js
```diff
@@ -42,20 +42,48 @@ const format = (event, options = { ignoreSelection: false }) => {
...
+    const textToTransform = isTransformingFile ? editor.getText() : selectedText;
 
-  const transformed = executePrettier(textToTransform, getFormatOptions());
-  if (!transformed) {
-    return;
-  }
+    const transformed = executePrettier(textToTransform, getFormatOptions());
+    if (!transformed) {
+      return;
+    }
```
robwise: it seems like there is some room to DRY this up, and the method is getting super-long now. Maybe we should extract method into something like `getTextToTransform(editor, isEmbedded)` or something?

lib/helpers.js
transcranial: With my own tests, this works for both manual formatting as well as format-on-save.

lib/helpers.js
transcranial: Right. It doesn't work for >1 script blocks, so I put in the guard for now. It could be because the buffer range gets misaligned or something similar. Worth investigating since >1 blocks should probably be supported. Btw, manually formatting two selections doesn't work currently either. Let me know if you have any ideas.

transcranial: Figured it out. Have to use `editor.backwardsScanInBufferRange` rather than `editor.scan` since transforming the text in the forward direction causes misalignment with the buffer range. Now I can get multiple script blocks working.

transcranial:
Refactored and cleaned up the code so that the common theme becomes applying transforms on ranges, which accounts for all the different scenarios. Good news is that formatting multiple selections now works, too.

lib/helpers.js
```diff
@@ -10,6 +10,11 @@ const PRETTIER_CORE_OPTIONS = [
   'bracketSpacing',
 ];
 
+const EMBEDDED_SCOPE = [
+  'text.html.vue',
+  'text.html.basic',
+];
```
erbridge: Shouldn't these be customizable?

transcranial: how so?

robwise: we could make them an option, but I'm not sure how used it will be? We can do that in a new PR if we want that.

Issue #45, opened 2017-02-11
Adding the ability to fail silently


oriSomething:
Current behave is that when I save a document that cannot be converted to AST, I get a huge popover of the failure of `prettier-atom`.
I think adding the ability that it will fail silently and won't be trigger the popover can be very helpful since, It's not exactly a pleasant UX.
Of course, I can try making a PR if needed.

Thanks!

robwise:
Yeah, great idea! Do you think you can try to tackle this one in a PR? I think we can just have a simple checkbox that lets you silence the errors? Does it do this only on format-on-save, or even if you invoke the command manually?

oriSomething:
- Yes, i'll try to PR this
- I though on a simple checkbox too
- If you invoke manually the command. It might won't be a good idea

erbridge:
I actually find those errors very helpful. They usually occur because of syntax issues I had missed. But perhaps they could highlight the error instead of making such a huge stack trace?

oriSomething:
@erbridge It's seems to me like a valid option. But, I don't think it should come instead of the "silent failure"

robwise:
Yeah we could have another option which is `errors include stack trace` or something? And that one applies to everything and not just the format on save?

robwise:
I'm going to close this for now, discussion of removing stack trace moved to https://github.com/jlongster/prettier-atom/issues/49

PR #46, opened 2017-02-13
new option: "Silent error on save"

oriSomething:
Continuation of #45 

Please review the code. I'm sure you might want to solve this problem differently

### DIFF
--- a/lib/helpers.js
+++ b/lib/helpers.js
@@ -10,10 +10,7 @@ const PRETTIER_CORE_OPTIONS = [
   'bracketSpacing',
 ];
 
-const EMBEDDED_SCOPE = [
-  'text.html.vue',
-  'text.html.basic',
-];
+const EMBEDDED_SCOPE = [ 'text.html.vue', 'text.html.basic' ];
 
 // helpers
 const getConfigOption = key => atom.config.get(`prettier-atom.${key}`);
@@ -24,22 +21,26 @@ const getFormatOptions = () => PRETTIER_CORE_OPTIONS.reduce(setConfigOption, {})
 
 const getCurrentScope = () => atom.workspace.getActiveTextEditor().getGrammar().scopeName;
 
-const executePrettier = (text, formatOptions) => {
+const executePrettier = (text, formatOptions, options = { ignoreErrors: false }) => {
   try {
     return prettier.format(text, formatOptions);
   } catch (e) {
-    const message = `prettier-atom: ${e.toString()}`;
-    const detail = e.stack.toString();
-    atom.notifications.addError(message, { detail, dismissable: true });
     console.log('Error executing Prettier:', e);
+
+    if (!options.ignoreErrors) {
+      const message = `prettier-atom: ${e.toString()}`;
+      const detail = e.stack.toString();
+      atom.notifications.addError(message, { detail, dismissable: true });
+    }
+
     return false;
   }
 };
 
-const transformRange = (editor, range) => {
+const transformRange = (editor, range, options = { ignoreErrors: false }) => {
   const cursorPositionPriorToFormat = editor.getCursorScreenPosition();
   const textToTransform = editor.getTextInBufferRange(range);
-  const transformed = executePrettier(textToTransform, getFormatOptions());
+  const transformed = executePrettier(textToTransform, getFormatOptions(), options);
   if (!transformed) {
     return;
   }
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -58,47 +58,54 @@
       "default": false,
       "order": 1
     },
+    "formatOnSaveSilentError": {
+      "title": "On format on Save, Don't show error message",
+      "description": "Don't show error if accour on format on save",
```
robwise: Typo? `accour`

package.json
```diff
@@ -58,47 +58,54 @@
       "default": false,
       "order": 1
     },
+    "formatOnSaveSilentError": {
+      "title": "On format on Save, Don't show error message",
```
robwise: There's some inconsistent capitalization going on here

lib/helpers.js
```diff
@@ -94,7 +93,12 @@ const formatOnSaveIfEnabled = () => {
     return;
   }
 
-  format(null, { ignoreSelection: true });
+  const formatOnSaveSilentError = getConfigOption('formatOnSaveSilentError');
+
+  format(null, {
+    ignoreSelection: true,
+    ignoreErrors: !formatOnSaveSilentError,
```
robwise: I would just inline the call to `getConfigOption`, or at least name the variablje `ignoreErrors` so you can just use property shorthand

lib/helpers.js
```diff
@@ -62,22 +61,22 @@ const format = (event, options = { ignoreSelection: false }) => {
...
     } else {
-      transformRange(editor, bufferRange);
+      transformRange(editor, bufferRange, { ignoreErrors: options.ignoreErrors });
     }
   } else {
     // transform selected ranges
-    editor.getSelectedBufferRanges().forEach(range => {
-      transformRange(editor, range);
+    editor.getSelectedBufferRanges().forEach((range) => {
+      transformRange(editor, range, { ignoreErrors: options.ignoreErrors });
```
robwise: we should just pass `options` instead of redefining a new object every time like this

lib/helpers.js
```diff
@@ -62,22 +61,22 @@ const format = (event, options = { ignoreSelection: false }) => {
...
         const { start, end } = iter.range;
-        const startModified = [start.row + 1, start.column];
+        const startModified = [ start.row + 1, start.column ];
         const range = new iter.range.constructor(startModified, end);
-        transformRange(editor, range);
+        transformRange(editor, range, { ignoreErrors: options.ignoreErrors });
       });
     } else {
-      transformRange(editor, bufferRange);
+      transformRange(editor, bufferRange, { ignoreErrors: options.ignoreErrors });
```
robwise: we should just pass `options` instead of redefining a new object every time like this

lib/helpers.js
```diff
@@ -62,22 +61,22 @@ const format = (event, options = { ignoreSelection: false }) => {
...
+      editor.backwardsScanInBufferRange(re, bufferRange, (iter) => {
         // Create new range with start row advanced by 1,
         // since we cannot use look-behind on variable-length starting
         // <script ...> tag
         const { start, end } = iter.range;
-        const startModified = [start.row + 1, start.column];
+        const startModified = [ start.row + 1, start.column ];
         const range = new iter.range.constructor(startModified, end);
-        transformRange(editor, range);
+        transformRange(editor, range, { ignoreErrors: options.ignoreErrors });
```
robwise: we should just pass `options` instead of redefining a new object every time like this

lib/helpers.js
```diff
@@ -24,22 +21,24 @@ const getFormatOptions = () => PRETTIER_CORE_OPTIONS.reduce(setConfigOption, {})
 
 const getCurrentScope = () => atom.workspace.getActiveTextEditor().getGrammar().scopeName;
 
-const executePrettier = (text, formatOptions) => {
+const executePrettier = (text, formatOptions, options = { ignoreErrors: false }) => {
   try {
     return prettier.format(text, formatOptions);
   } catch (e) {
     const message = `prettier-atom: ${e.toString()}`;
```
robwise: we should use a guard clause instead that way it's clear what code goes with what:

```js
  console.log('Error executing Prettier:', e);

  if (!options.ignoreErrors) {
    const message = `prettier-atom: ${e.toString()}`;
    const detail = e.stack.toString();
    atom.notifications.addError(message, { detail, dismissable: true });
  }
  
  return false;
}
```

lib/helpers.js
oriSomething: But the options of `format` contains options which aren't related to `transformRange`. I prefer to pass to a function only properties that are relevant to that function

oriSomething:
By the way tell when code is ready so I'll rebase history it before merge

lib/helpers.js
robwise: ok fair enough

oriSomething:
Done

PR #51, opened 2017-02-15
Update to use proper options for 0.17.1 of Prettier

skevy:
Changes "useFlowParser" to "parser" (now a string enum), and adds "jsxBracketSameLine".

### DIFF
--- a/package.json
+++ b/package.json
@@ -102,7 +102,7 @@
     },
     "jsxBracketSameLine": {
       "title": "JSX Bracket Same Line",
-      "description": "Put > on the last line in JSX",
+      "description": "Put > on the same line in JSX",
       "type": "boolean",
       "default": false,
       "order": 8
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -100,13 +100,21 @@
       "default": true,
       "order": 7
     },
-    "useFlowParser": {
-      "title": "Use Flow Parser",
-      "description": "Use the [flow](https://github.com/facebook/flow) parser instead of [babylon](https://github.com/babel/babylon).",
+    "jsxBracketSameLine": {
+      "title": "JSX Bracket Same Line",
+      "description": "Put > on the last line in JSX",
```
robwise: The title and the description seem to represent opposite states. Does `false` mean it puts it on the same line or the last line?

skevy: I thought I copied from the prettier docs...but you're right. It should be "Put > on the same line in JSX". Will fix.

robwise:
Fixes https://github.com/jlongster/prettier-atom/issues/52

PR #76, opened 2017-03-06
fix($config): Unset old config options causing errors

### LINKED ISSUE #72: trailingComma validation error
Version: 0.21.2

My config.cson has:

```
  "prettier-atom":
    prettierOptions: {}
    singleQuote: true
    trailingComma: false
    useEslint: true
```

Error:

```
prettier-atom: ● Validation Error:

Option "trailingComma" must be of type: boolean but instead received: string

Example: { "trailingComma": false }
● Validation Error:
  Option "trailingComma" must be of type:
    boolean
  but instead received:
    string
  Example:
  {
    "trailingComma": false
  }
```

robwise:
It seems that Atom may have a bug that is causing errors for some users. When we moved to a new
config schema, some users still have the old configuration options still set. This is preventing
them from loading prettier in some cases, because the options have incorrect values. It's difficult
to tell what is going on here and this fix may not work.

Fixes #72

### DIFF
--- a/dist/main.js
+++ b/dist/main.js
@@ -58,6 +58,8 @@ var activate = function activate() {
   subscriptions.add(atom.config.observe('prettier-atom.useEslint', function () {
     return lazyWarnAboutLinterEslintFixOnSave();
   }));
+  atom.config.unset('prettier-atom.singleQuote');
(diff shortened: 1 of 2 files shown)


aaronshaf:
Thanks!

PR #81, opened 2017-03-07
feat(whitelist-globs): add ability to whitelist globs

### LINKED ISSUE #62: Whitelist Globs
Feature request from @kentcdodds to add an option for whitelisting certain globs. I'm thinking that these would take priority over both the blacklist and the scopes, so we check the whitelist first, and if the globs match the file, we go straight to executing.

robwise:
Whitelist globs will take precedence over any scopes and excluded globs. For more information on how
globs are matched, see the documentation at https://github.com/isaacs/minimatch. We are not using
any special options from the minimatch library.

Closes #62

### DIFF
--- a/dist/config-schema.json
+++ b/dist/config-schema.json
@@ -52,10 +52,17 @@
       },
       "excludedGlobs": {
         "title": "Exclude (list of globs)",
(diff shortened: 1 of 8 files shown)


kentcdodds:
Yay!

PR #83, opened 2017-03-09
feat(whitelist): force whitelist if provided

kentcdodds:
**What**: Update the logic of whitelist to be required if provided

**Why**: The use case of a whitelist is for people who have various projects on their machine. Some of them they are able to use `prettier-atom` on and others they are not. It's often easier to maintain a list of projects they're allowed to use `prettier-atom` on (whitelist) than it is to maintain a list of projects they are _not_ allowed to use `prettier-atom` on (blacklist). For these users, we say that if you do provide a whitelist, then the files _must_ match the whitelist or they will _not_ be formatted on save.

**How**:
- Add method `isWhitelistProvided` to `helpers`
- Add logic in `formatOnSaveIfAppropriate` for skipping if the whitelist is provided and the file is not whitelisted
- Add note in config schema about this behavior

### DIFF
--- a/src/formatOnSave.test.js
+++ b/src/formatOnSave.test.js
@@ -38,6 +38,8 @@ test('it executes prettier on buffer range', () => {
 
 test('it executes prettier on buffer range if file is whitelisted regardless of exclusions or scopes', () => {
   // $FlowFixMe
+  helpers.isWhitelistProvided.mockImplementation(() => true);
+  // $FlowFixMe
   helpers.isInScope.mockImplementation(() => false);
   // $FlowFixMe
   helpers.isFilePathExcluded.mockImplementation(() => true);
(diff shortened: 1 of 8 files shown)


src/formatOnSave.test.js
```diff
@@ -49,6 +51,21 @@ test('it executes prettier on buffer range if file is whitelisted regardless of
   expect(executePrettierOnBufferRange).toHaveBeenCalledWith(editor, rangeFixture);
 });
 
+test('it does not execute prettier if whitelist is provided and file is not whitelisted', () => {
```
kentcdodds: This is the key here. The added code is to support this test

kentcdodds:
I should also note that I don't find the current incarnation of whitelist very useful as it only really allows you to whitelist things that would otherwise be blacklisted. A use case like that would be if I'm in a project that I've blacklisted, but there's a section of the project that I want to whitelist. I can see that being handy in some small cases, but not many. By adding these additional semantics, it makes the feature much more useful (while still maintaining the current use case).

src/formatOnSave.test.js
```diff
@@ -49,6 +51,21 @@ test('it executes prettier on buffer range if file is whitelisted regardless of
...
+  // $FlowFixMe
+  helpers.isInScope.mockImplementation(() => true);
+  // $FlowFixMe
+  helpers.isFilePathExcluded.mockImplementation(() => false);
+  // $FlowFixMe
+  helpers.isFilePathWhitelisted.mockImplementation(() => false);
+
+  formatOnSave(editor, filePathFixture);
+
+  expect(executePrettierOnBufferRange).toHaveBeenCalledTimes(0);
```
robwise: J/C, do you prefer this over `.not.toHaveBeenCalled()`?

kentcdodds: Great question. I do prefer it because I actually like to have that assertion for every one of my mock functions I care about. Rather than just assuming that a function was called once, I like to be explicit about how many times it's called. So I prefer `toHaveBeenCalledTimes` over `toHaveBeenCalled` in general, and I extend that to when it's not called as well :)

kentcdodds:
> I think I just got confused about the requirements. I agree this makes more sense.

It's a confusing thing, no worries :)

PR #102, opened 2017-03-25
note added for app restart after package installation

da-vaibhav:


kentcdodds:
I'm going to go ahead and let @robwise review this as well because there may be more information he knows that I don't. Thanks!

Issue #59, opened 2017-03-02
Select bin file from repo


note89:
When running this plugin you are stuck with the version bundled with the package. 
Would be great if one could use the version installed in `node_modules`.

Now i get a diff from running prettier via the command line and the atom package

robwise:
Hmm, this is a great point. I am currently doing a big refactor but I will definitely plan on implementing this somehow.

n1k0:
This is a really important point, as many different versions may existing across one's repositories.

charypar:
@robwise Would you be interested in a PR making the change?

robwise:
@charypar Sure! Do you want to take a stab at it? I figure we should prefer to look for the user's version of it first, and if we don't find it, then use our own packaged version. The only trick is that what you're looking for depends on whether they have the eslint integration turned on or not (look for `prettier` if not, look for `prettier-eslint` if so).

charypar:
That was my thinking too. I'll have a look at the complexity around `prettier-eslint` and try to have a first stab at it by the end of this week. Thanks for the pointers!

charypar:
@robwise I had a go and got it working using a dynamic `require`, but both eslint and flow are complaining about it and I'm thinking they are probably right - dynamic require does seem a bit iffy. 

Would it be better to switch to using prettier through its CLI (similar to how linter-eslint works)? Prettier-eslint doesn't have a CLI, but as far as I can tell, it already does a local lookup (https://github.com/prettier/prettier-eslint/blob/master/src/index.js#L45), so it should be possible to leave that as is.

I thought I'd ask your opinion before proceeding any further.

robwise:
> Prettier-eslint doesn't have a CLI

@charypar It does, check this out: https://github.com/prettier/prettier-eslint-cli

> as far as I can tell, it already does a local lookup 

It looks like prettier-eslint does a local lookup of eslint and prettier, but do you think we should also do a local lookup of prettier-eslint itself?

> using a dynamic require, but both eslint and flow are complaining about it and I'm thinking they are probably right - dynamic require does seem a bit iffy.

Well, I'm thinking everyone is going to have the normal package installed, but not necessarily everyone will have the CLI installed? I was thinking we would recursively search up the file tree until we find a node_modules folder, and then check for it in there, and then call it? I think this is the require relative idea that you've done?

n1k0:
> Well, I'm thinking everyone is going to have the normal package installed, but not necessarily everyone will have the CLI installed?

[...]

Issue #43, opened 2017-02-07
Only format on save if prettier is in the dependencies


mxstbr:
I have projects that use `prettier` and adhere to the style standard, but I also have projects that neither use `prettier` nor adhere to the style. (they enforce some other styleguide with `eslint`)

It'd be great if the "Format on save" option only took effect in projects that have `prettier` in their `package.json` or something like that!

robwise:
I was thinking in https://github.com/jlongster/prettier-atom/issues/41 that it might be good to do a blacklist glob, that might be better because that would cover cases where people still want to use prettier for their files even though the project might not have it listed in the dependencies.

mxstbr:
> it might be good to do a blacklist glob

That seems very manual. I create new projects almost everyday, and updating a custom glob every time based on "do I want to use prettier or not" seems pretty tedious. I'm not sure I'd even enable that option just based on the fact that I'll probably forget about it.

darahak:
Why not both?
I think this could be fixed by a `Use global installation` setting, like in the `linter-eslint` package.
And the blacklist could solve [prettier#787](https://github.com/prettier/prettier/issues/787) for example

blainekasten:
I'm with @mxstbr here. I came here to raise this issue. Glad to see I'm not the only one facing this. I personally do not create a new repo every day, but I'm even just thinking organizationally as I try to introduce prettier to my company, asking every person in the company to set up their blacklist creates friction points to adoption.

robwise:
Sorry if I've been unclear on this, we are definitely going to get this addressed. Right now the plan is we are going to look for prettier/prettier-eslint (depending on your eslint integration setting), and if we don't find it, we will have an option in the settings where you can specify if you want us to fall back to the included packaged version or just not do anything. https://github.com/jlongster/prettier-atom/issues/59#issuecomment-287848381

PR #104, opened 2017-03-27
Prefer local prettier over bundled version

charypar:
Following on the conversation in #59, this is the first cut of the implementation.

Calling prettier through the CLI works fine, but does introduce a short but noticeable lag (~300 ms on a MacBook Pro, so probably even more significant elsewhere). This is mentioned as a potential issue in https://github.com/prettier/prettier/issues/918. 

@robwise Is this a significant enough performance penalty to go back to dynamic require until jsonrpc or something similar is introduced in prettier?

I still need to implement the configuration option requested in #43.

# to do

- [x] Prefer local prettier over bundled version
- [x] Error handling
- [ ] ~~Add configuration to only use local prettier and do nothing if there isn't one~~


src/helpers.js
```diff
@@ -105,6 +125,11 @@ const getPrettierOptions = (editor: TextEditor) => ({
   jsxBracketSameLine: getPrettierOption('jsxBracketSameLine'),
 });
 
+const getPrettierCliOptions = (options: Object) =>
+  Object.keys(options)
+    .reduce((cliOptions, option) => [...cliOptions, `${CLI_OPTIONS[option]} ${options[option]}`], [])
+    .join(' ');
```
robwise: It seems like you're using a `reduce` to imitate a `map`? I think you could just use a map directly:

```js
const getPrettierCliOptions = (prettierOptions: {}) => 
 Object.keys(prettierOptions).map(option => `${CLI_OPTIONS[option]} ${prettierOptions[option]`).join(' ');
```

or you could stay with reduce and skip the array/join:

```js
const getPrettierCliOptions = (prettierOptions: {}) => 
 Object.keys(prettierOptions).reduce((acc, option) => `${acc} ${CLI_OPTIONS[option]} ${prettierOptions[option]`);
```

Also, I think we maybe should just take the `editor` as an argument and call `prettierOptions` ourselves instead of making the caller do it, that way the caller is less coupled:

```js
const getPrettierCliOptions = (editor: TextEditor) => 
 Object.keys(getPrettierOptions(editor)).reduce((acc, option) => `${acc} ${CLI_OPTIONS[option]} ${prettierOptions[option]`);
```

src/helpers.test.js
```diff
@@ -312,6 +338,36 @@ describe('getPrettierOptions', () => {
...
+  test('translates a single boolean option', () => {
+    const expected = '--single-quote true';
+    const actual = getPrettierCliOptions({ singleQuote: true });
+
+    expect(actual).toEqual(expected);
+  });
+
+  test('translates complex options', () => {
+    const expected = '--single-quote true --print-width 100 --trailing-comma es5';
+    const actual = getPrettierCliOptions({ singleQuote: true, printWidth: 100, trailingComma: 'es5' });
```
robwise: this is nit-picky, but can we reverse the order of `expected` and `actual` to match above?

src/helpers.test.js
```diff
@@ -312,6 +338,36 @@ describe('getPrettierOptions', () => {
   });
 });
 
+describe('getPrettierCliOptions', () => {
+  test('translates empty options', () => {
+    const expected = '';
+    const actual = getPrettierCliOptions({});
+
+    expect(actual).toEqual(expected);
+  });
```
robwise: I don't think this case should ever occur, should it? We're always going to retrieve all options, right?

robwise:
I just read about [this](https://github.com/gustavnikolaj/linter-js-standard-engine/blob/master/lib/worker.js) in the prettier-atom [RFC for linter integration](https://github.com/prettier/prettier/issues/918#issuecomment-289680645). I'm not totally sure I understand all of this properly.

src/helpers.test.js
charypar: Happy to remove that, that's a relic of my test-driving the implementation :)

src/helpers.test.js
charypar: Sure.

src/helpers.js
charypar: Ohh 🤦‍♂️  this is what happens when I write code in the evening. This used to concatenate strings like your second suggestion, that's why `reduce`, then I changed my mind to do an array and a join and didn't notice it's now just a map.

Happy to change it to take `editor` instead of options.

charypar:
I thought about this some more over night and I think we could solve the performance problem by having the `prettier` process running in the background and streaming input/output. That way, we'd only pay the penalty once when starting the process and from then on it would be quick. If I imagine it correctly that is.

I also realised I haven't done any error handling for the CLI call. Added it in the to do list at the top.

charypar:
> Hmm, good question here. If it's just the initial/first time, then I think it's fine. If it's every time, I'm not as sure, 300ms sounds like a lot in some contexts, in others it's not even detectable. I'm not really sure on this.

It is every time now. I'm not very happy with it as is.

> Should we leave the option (whether it should fallback or not) for a new PR and merge this?

Up to you, happy to do either.

robwise:
> I also realised I haven't done any error handling for the CLI call. Added it in the to do list at the top.

Yeah, now that I think of it, it may be very difficult to parse the stack trace/error that comes out of the CLI. When you combine that with the performance hit and the complication of needing to run it in the background (which is sort of begging for memory bloat if not leaks IMO), maybe it's better to go with programatically invoking it instead of using the CLI?

robwise:
> Up to you, happy to do either.

In that case, can you do a separate PR when you get to that step? That way the PRs are nice and tight! BTW, anytime you want to pair or just if you feel like it's too much work and want to hand it off, let me know, I really appreciate all of this help.

charypar:
> Yeah, now that I think of it, it may be very difficult to parse the stack trace/error that comes out of the CLI. When you combine that with the performance hit and the complication of needing to run it in the background (which is sort of begging for memory bloat if not leaks IMO), maybe it's better to go with programatically invoking it instead of using the CLI?

I think running it as a child process, while it requires some work, will work best - no dynamic requires, no startup penalty and we can make sure to only ever have one instance running (closing the previous one if there is one). 

You're right the error output could be difficult to parse, but as it stands, I don't think the error reporting is parsing the message at all, does it? It's just printed into an alert.

I'll give it a couple more hours tonight and if it gets crazy, lets go with a dynamic `require`. How does that sound?

> In that case, can you do a separate PR when you get to that step?

Sure, will open as a separate PR.

robwise:
Sounds good to me!

charypar:
I looked into whether the streaming will work and I don't think it will without changes to prettier itself - it currently waits for `stdin` to close before attempting to format and returning any output.

So that settles it - it's got to be a dynamic `require` based on how eslint / prettier-eslint does things. Oh well :)

robwise:
@charypar Should I pick this one up?

charypar:
@robwise Oh, thanks for the reminder. Past couple weeks were busy, but I've got time tonight, I'll make the changes and push them in a couple hours.

charypar:
@robwise Ok, I think this is ready to merge, if you're happy with it. 

Special error handling is no longer needed, errors can be handled the same way for both bundled and local prettier.

Speaking of errors, would you be interested in a PR that would show parse errors from prettier inline using `atom-linter`, like eslint does?

src/executePrettier.js
```diff
@@ -37,7 +47,15 @@ const executePrettier = (editor, text) => {
...
-    return prettier.format(text, getPrettierOptions(editor));
+
+    const prettierOptions = getPrettierOptions(editor);
+    const localPrettier = getLocalPrettierPath(getCurrentFilePath(editor));
+
+    if (!localPrettier) {
+      return prettier.format(text, prettierOptions);
+    }
+
+    return getLocalPrettier(localPrettier).format(text, prettierOptions);
```
robwise: Code style comment: I think there's more than one responsibility going on in this function. This function is supposed to be responsible for executing prettier, but now it also has the responsibility for figuring out which prettier to use and where to get it.

I think it would be better, therefore, if we made a helper method simply called `getPrettier` that would handle all of this internally. This would also make mocking much easier because we can just mock `getPrettier`.

src/helpers.js
```diff
@@ -22,6 +22,16 @@ const getDirFromFilePath = (filePath: FilePath): FilePath => path.parse(filePath
 const getNearestEslintignorePath = (filePath: FilePath): ?FilePath =>
   findCached(getDirFromFilePath(filePath), '.eslintignore');
 
+const getLocalPrettierPath = (filePath: ?FilePath): ?FilePath => {
+  if (!filePath) return null;
+
+  const indexPath = path.join('node_modules', 'prettier', 'index.js');
+  const dirPath = getDirFromFilePath(filePath);
+  if (!dirPath) return null;
```
robwise: Is this guard clause possible to hit? If not, you can go FP and avoid state:

```js
const getLocalPrettierPath = (filePath: ?FilePath): ?FilePath => 
  filePath ?
    findCached(getDirFromFilePath(filePath), path.join('node_modules', 'prettier', 'index.js'))
    : null;
```

Also, see my comment in `executePrettier.js`. There's really no reason to not just give the `prettier` instance itself instead of making `executePrettier` do it.

src/helpers.test.js
```diff
@@ -96,6 +97,43 @@ describe('getCurrentFilePath', () => {
...
+
+    atomLinter.findCached.mockImplementation(() => prettierLib);
+    const filePath = path.join(__dirname, '..', 'tests', 'fixtures', 'sourceFile.js');
+
+    getLocalPrettierPath(filePath);
+    const expectedDir = path.join(__dirname, '..', 'tests', 'fixtures');
+    const expectedLib = path.join('node_modules', 'prettier', 'index.js');
+
+    expect(atomLinter.findCached).toHaveBeenCalledWith(expectedDir, expectedLib);
+  });
```
robwise: I think this test seems to pretty much be checking the same thing as the one above? Should we just keep one?

src/executePrettier.js
charypar: Sure, that makes more sense, we can then just past the prettier instance to use to the `executePrettier` function. 

We could even pass prettier down the execute stack from `format` / `formatOnSave`, like the `editor`.

src/helpers.js
charypar: Yep, makes sense, ternary it is. The reason to split the path/lookup logic from the actual 
requiring is it makes it easier to test in isolation.

src/executePrettier.js
robwise: I don't know if I'd do the `format` thing, then you're putting in the logic in two different places that otherwise had no need to know anything about how `executePrettier` works?

src/helpers.test.js
charypar: One is checking we get the right thing back given `findCached` returns it, the other checks we call `findCached` with the right arguments.

I had them in one test at first, but this way they can fail independently. Happy to merge them together if you'd prefer.

robwise: > One is checking we get the right thing back given findCached returns it, the other checks we call findCached with the right arguments.

Yeah although I'd argue the first one is what we need the function to do, and the second one is just an implementation detail, so I think we can drop that one.

src/executePrettier.js
charypar: You could move it all the way up to `main` to do it in one place and pass it through both `format` functions, I suppose? 

Although I guess we might as well hide the problem in the `helpers` by wrapping it in one function :)

charypar:
I'll make those changes tonight, should be quick.

charypar:
@robwise This should address all the review comments. Please let me know if you want to change anything else.

robwise:
Awesome! Thanks!

PR #113, opened 2017-04-02
fix(prettier): Don't alter text buffer if file is already pretty

lgeiger:
This PR brings two improvements, it:
- removes unnecessary glitches that happen when the content of the editor is overwritten
- preserves the history, since it removes unnecessary text buffer changes

### DIFF
--- a/src/executePrettier.js
+++ b/src/executePrettier.js
@@ -47,7 +47,8 @@ const executePrettierOnBufferRange = (editor: TextEditor, bufferRange: Range) =>
   const textToTransform = editor.getTextInBufferRange(bufferRange);
   const transformed = executePrettier(editor, textToTransform);
 
-  if (!transformed) return;
+  const isTextUnchanged = transformed === textToTransform;
+  if (!transformed || isTextUnchanged) return;
 
   editor.setTextInBufferRange(bufferRange, transformed);
   editor.setCursorScreenPosition(cursorPositionPriorToFormat);
--- a/src/executePrettier.test.js
+++ b/src/executePrettier.test.js
@@ -77,6 +77,24 @@ describe('executePrettierOnBufferRange()', () => {
     });
   });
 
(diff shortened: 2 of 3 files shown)


src/executePrettier.js
```diff
@@ -47,7 +47,7 @@ const executePrettierOnBufferRange = (editor: TextEditor, bufferRange: Range) =>
   const textToTransform = editor.getTextInBufferRange(bufferRange);
   const transformed = executePrettier(editor, textToTransform);
 
-  if (!transformed) return;
+  if (!transformed || transformed === textToTransform) return;
```
robwise: Now that this conditional is a bit more complex, can we put this in a const to self-document (just for readability later on)?:

```js
const isTextUnchanged = transformed === textToTransform;
if (!transformed || isTextUnchanged) return;
```

src/executePrettier.test.js
```diff
@@ -77,6 +77,25 @@ describe('executePrettierOnBufferRange()', () => {
     });
   });
 
+  describe('when file is already pretty', () => {
+    beforeEach(() => {
+      prettier.format.mockImplementation(() => 'untransformed text');
+      prettierEslint.mockImplementation(() => 'untransformed text');
```
robwise: Can we remove this line? It should only be using `prettierEslint` if you mocked out `helpers.shouldUseEslint.mockImplementation(() => true);`?

src/executePrettier.test.js
```diff
@@ -77,6 +77,25 @@ describe('executePrettierOnBufferRange()', () => {
     });
   });
 
+  describe('when file is already pretty', () => {
```
robwise: We should probably just say `when text in buffer range is already pretty` or something similar since we can format selections as well?

lgeiger:
👍 Amended the fixes

robwise:
Wow!! That was fast thanks for the help!

PR #123, opened 2017-04-14
fix(embedded-scripts): skip single-line script tags in html files

transcranial:


transcranial:
@robwise Done. Regarding single-line formatting use-cases, I can't really think of any right now.

robwise:
Roger that, thanks!

PR #130, opened 2017-04-16
Enable context menu entry for JSX files

darahak:
This time, with commitizen.

Context menus didn't include the `Format` action for files that use `Babel` or `JavaScript with JSX` for syntax highlight (from the `language-babel` package).


robwise:
Thanks!

PR #132, opened 2017-04-17
Add a prettier:debug command

darahak:
Users often create issues without providing enough details to help maintainers understand their problems.

This adds a `prettier:debug` command that displays info about the current installation of `prettier-atom`: versions, detailed configuration...
Got the idea from [linter-eslint](https://github.com/AtomLinter/linter-eslint).

Here's how it looks:
prettier:debug example

I also added an issue template to offer some guidance, including using this new command (`CONTRIBUTING.md` doesn't say anything about creating issues).


darahak:
This includes the build result from #123 apparently. I didn't notice before committing.

src/helpers.test.js
```diff
@@ -405,3 +406,50 @@ describe('runLinter()', () => {
...
+    const mockGet = jest.fn(() => expectedConfig);
+    atom = {
+      getVersion: jest.fn(),
+      config: { get: mockGet },
+    };
+
+    const info = getDebugInfo();
+
+    expect(info.prettierAtomConfig).toMatchObject(expectedConfig);
+  });
```
robwise: Can we combine this test in the one above? Also, I think we can use a simple string or something to make sure `prettierAtomConfig` is getting set to the right value and that would make it a bit shorter:

```js
const expectedConfig = 'atom prettier config';

// ... 

expect(info.prettierAtomConfig).toEqual(expectedConfig);
```

src/helpers.js
```diff
@@ -141,6 +144,19 @@ const runLinter = (editor: TextEditor) =>
     ? atom.commands.dispatch(atom.views.getView(editor), LINTER_LINT_COMMAND)
     : undefined);
 
+const getDebugInfo = () => {
+  const prettierAtomInfo = readPkg.sync();
+  const prettierInfo = readPkg.sync(getDepPath('prettier'));
+  const prettierESLintInfo = readPkg.sync(getDepPath('prettier-eslint'));
```
robwise: can we inline these in the object being returned, that way we can avoid having any local state?

src/helpers.js
```diff
@@ -77,6 +78,8 @@ const isLinterLintCommandDefined = (editor: TextEditor) =>
     .findCommands({ target: atom.views.getView(editor) })
     .some(command => command.name === LINTER_LINT_COMMAND);
 
+const getDepPath = dep => path.join(__dirname, '../node_modules', dep);
```
robwise: can you add a flow annotation for what `dep` is supposed to be (I assume `string` or something, right?)

darahak:
@robwise Thanks for your feedback! I'll address your points tomorrow. It's time to sleep now 😄

darahak:
@robwise All done!

Issue #117, opened 2017-04-13
A way to enable/disable prettier via the command menu


tleunen:
Currently, the only shortcut we have is `format`. But it would be great to have a way to quickly enable/disable prettier.

Asking because I switch from project to another and not all of them use prettier so I have to enable or disable it manually.

![image](https://cloud.githubusercontent.com/assets/1972567/25010598/77f016ae-201f-11e7-9657-a68b139d4903.png)

robwise:
Yeah, I think this is a good idea!

darahak:
Do you want a toggle command just for `fix-on-save` or everything including `format`? (the shortcut would be inactive)

robwise:
I guess just fix-on-save?

robwise:
@darahak are you taking this one or just discussing? I can assign you if you want to take a stab at it?

tleunen:
Yep, talking about this thing:
![image](https://cloud.githubusercontent.com/assets/1972567/25300392/fbaf4a88-26c2-11e7-834e-19d77269bd56.png)

darahak:
@robwise Yup, I'll take this one. Since I already added one command, this seems like a natural follow-up.

FezVrasta:
I think it would make sense to have a dotfile on the project directory to enable it.

Doing so I can keep it always on and have it active only in projects that explicitly states they make use of it.

Alternatively you may read the package.json and enable it only if it is listed as dependency.

PR #142, opened 2017-04-25
Toggle command for "Format on Save" option

darahak:
Fixes https://github.com/prettier/prettier-atom/issues/117.

The new command is defined as `prettier:toggle-format-on-save`.

I also wanted to add a visual indicator for the command, because it's not practical when you don't know if the option is already enabled or not.

I tried to make something in the status bar with Prettier's logo, but it looked terrible with small dimensions and it didn't fit with the other tiles.
I ended up just adding a "Prettier" label:

<details>

**Enabled**

**Disabled**

</details>

### DIFF
--- a/src/statusTile.js
+++ b/src/statusTile.js
@@ -25,6 +25,8 @@ const updateStatusTile = (disposable: Atom$Disposable, element: HTMLElement) => 
   }
   tooltip = atom.tooltips.add(element, { title: `Format on Save: ${getFormatOnSaveStatus()}` });
   disposable.add(tooltip);
+
+  return tooltip;
 };
 
 module.exports = {
--- a/src/statusTile.js
+++ b/src/statusTile.js
@@ -16,13 +16,18 @@ const createStatusTile = () => {
   return element;
 };
 
+const disposeTooltip = () => {
+  if (tooltip) {
+    tooltip.dispose();
+  }
+};
+
 const updateStatusTile = (disposable: Atom$Disposable, element: HTMLElement) => {
   // eslint-disable-next-line no-param-reassign
   element.dataset.formatOnSave = getFormatOnSaveStatus();
 
-  if (tooltip) {
-    tooltip.dispose();
-  }
+  disposeTooltip();
+
   tooltip = atom.tooltips.add(element, { title: `Format on Save: ${getFormatOnSaveStatus()}` });
   disposable.add(tooltip);
 
@@ -32,4 +37,5 @@ const updateStatusTile = (disposable: Atom$Disposable, element: HTMLElement) => 
 module.exports = {
   createStatusTile,
   updateStatusTile,
+  disposeTooltip,
 };
--- a/src/statusTile.js
+++ b/src/statusTile.js
@@ -1,0 +1,33 @@
+// @flow
+const { getConfigOption } = require('./helpers');
+
+let tooltip: Atom$Disposable = null;
+
+const getFormatOnSaveStatus = () => getConfigOption('formatOnSaveOptions.enabled') ? 'enabled' : 'disabled';
+
+const createStatusTile = () => {
+  const element = document.createElement('div');
+  element.classList.add('prettier-atom-status-tile');
+  element.classList.add('inline-block');
+  element.dataset.formatOnSave = getFormatOnSaveStatus();
+
+  element.appendChild(document.createTextNode('Prettier'));
+
+  return element;
+};
+
+const updateStatusTile = (disposable: Atom$Disposable, element: HTMLElement) => {
+  // eslint-disable-next-line no-param-reassign
+  element.dataset.formatOnSave = getFormatOnSaveStatus();
+
+  if (tooltip) {
+    tooltip.dispose();
+  }
+  tooltip = atom.tooltips.add(element, { title: `Format on Save: ${getFormatOnSaveStatus()}` });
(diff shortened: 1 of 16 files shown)


RoM4iK:
I think a visual indicator is pretty cool, but it's better to provide a way to disable it in options.

src/toggleFormatOnSave.test.js
```diff
@@ -0,0 +1,18 @@
...
+      get: jest.fn(() => enabled),
+      set: jest.fn((key, value) => (enabled = value)),
+    },
+  };
+
+  toggleFormatOnSave();
+  expect(getConfigOption('...')).toBe(true);
+  toggleFormatOnSave();
+  expect(getConfigOption('...')).toBe(false);
+});
```
robwise: Can we write this as two tests? Something like:

```js
test('it sets formatOnSaveOptions.enabled to false if it was true', () => {
  atom = {
    config: {
      get: jest.fn(() => true),
      set: jest.fn(),
    }
  }

  toggleFormatOnSave();

  expect(atom.config.set).toHaveBeenCalledWith('atom-prettier.formatOnSaveOptions.enabled', false);
}

test('it sets formatOnSaveOptions.enabled to true if it was false', () => {
  atom = {
    config: {
      get: jest.fn(() => false),
      set: jest.fn(),
    }
  }

  toggleFormatOnSave();

  expect(atom.config.set).toHaveBeenCalledWith('atom-prettier.formatOnSaveOptions.enabled', true);
}
```

src/toggleFormatOnSave.js
```diff
@@ -0,0 +1,9 @@
+// @flow
+const { getConfigOption, setConfigOption } = require('./helpers');
+
+const toggleFormatOnSave = () => {
+  const key = 'formatOnSaveOptions.enabled';
```
robwise: can we move this outside of the function and uppercase since it's a constant?

src/statusTile.test.js
```diff
@@ -0,0 +1,36 @@
...
+
+  const disposable = { add: jest.fn() };
+  const tooltip = updateStatusTile(disposable, div);
+  expect(tooltip.prop).toBe('...');
+  expect(disposable.add).toHaveBeenCalled();
+
+  tooltip.dispose = jest.fn();
+  updateStatusTile(disposable, div);
+  expect(tooltip.dispose).toHaveBeenCalled();
+});
```
robwise: It seems we are testing multiple different functions in the same test? I'm not sure if this would work, but can we split them so that it's a bit easier to see what each one does, and also so that each function is tested in isolation of the other one? Something like this, although it requires creating and mocking out a `disposeCurrentTooltip` function or something:

```js
beforeEach(() => {
  atom = {
    config: { get: jest.fn(() => true) },
    tooltips: { add: jest.fn() },
  };
});

describe('createStatusTile()', () => {
  test('it creates a div with "Prettier" and tooltip status indicating formatOnSave status', () => {
    global.document = {
      createElement: jest.fn(() => ({
        classList: {
          add: jest.fn(),
        },
        dataset: {},
        appendChild: jest.fn(),
      })),
      createTextNode: jest.fn(arg => arg),
    };
    
    const div = createStatusTile();
    
    expect(div.dataset.formatOnSave).toBe('enabled');
    expect(div.classList.add).toHaveBeenCalledWith('prettier-atom-status-tile');
    expect(div.appendChild).toHaveBeenCalledWith('Prettier');
  });
});

describe('updateStatusTile()', () => {
  it('disposes previous tooltip and adds a new one with the current formatOnSave status', () => {
    const div = {};
    const disposable = { add: jest.fn() };

    updateStatusTile(disposable, div);

    expect(disposeCurrentTooltip).toHaveBeenCalled(); // assuming you've made this an exported function and mocked it
    expect(atom.tooltips.add).toHaveBeenCalledWith(div, {title: 'Format on Save: enabled'});
  });
});
```

src/statusTile.js
```diff
@@ -0,0 +1,35 @@
...
+
+  return element;
+};
+
+const updateStatusTile = (disposable: Atom$Disposable, element: HTMLElement) => {
+  // eslint-disable-next-line no-param-reassign
+  element.dataset.formatOnSave = getFormatOnSaveStatus();
+
+  if (tooltip) {
+    tooltip.dispose();
```
robwise: Does the tooltip also get disposed if prettier-atom is deactivated? I think we might want to export a function from this module called something like `deactivateTooltip` and then make sure we call that in `deactivate`? This would also allow you to mock this function with Jest when testing, so that we don't have to call `createStatusTile` in a `udpateStatusTile` test (which would prevent overlapping tests).

darahak:
@robwise I don't have much time right now, but I'll definitely address your requests.
Regarding the option for the status bar display, I'd rather make a separate PR for it.
I don't want this one to get too big :smile:

robwise:
@darahak Take your time, I appreciate any help at all. Another thing we can do if you want is just to merge it how it is and then I can make the changes myself later?

darahak:
@robwise Thanks for the proposal.
I'd like to do it myself if you don't mind. It's valuable to me, as a learning process.

darahak:
@robwise I couldn't figure out how to properly mock and test `disposeTooltip` as an exported function.
I did it differently, but feel free to change it.

robwise:
Okay roger that, I think it looks great now!

PR #143, opened 2017-04-26
feat(config): add .editorconfig support

### LINKED ISSUE #90: editorconfig integration
Hi, thanks for a very neat project! Looking forward to using it more.

I noted quickly however that the `tabWidth` setting doesn't seem to pick up the value as I've configured it in my `.editorconfig` file (see http://editorconfig.org/ for more details on that project). Would be great if the `tabWidth: auto` setting could pick up the value from there – that would make for a very portable workflow!

RoM4iK:
Support the loading of .editorconfig configuration over plugin global config

fixes #90


src/helpers.js
```diff
@@ -69,7 +70,7 @@ const someGlobsMatchFilePath = (globs: Globs, filePath: FilePath) => ignore().ad
 const getAtomTabLength = (editor: TextEditor) =>
   atom.config.get('editor.tabLength', { scope: editor.getLastCursor().getScopeDescriptor() });
 
-const useAtomTabLengthIfAuto = (editor, tabLength) =>
+const tabLengthOption = (editor, tabLength) =>
```
robwise: Why this change?

src/helpers.test.js
```diff
@@ -375,6 +378,45 @@ describe('getPrettierEslintOptions', () => {
...
+      trim_trailing_whitespace: true,
+      insert_final_newline: true,
+      tab_width: 2,
+      max_line_length: 100,
+    }));
+    const expected = {
+      tabWidth: 2,
+      printWidth: 100,
+    };
+    expect(getEditorConfigOptions('')).toEqual(expected);
```
robwise: can we use the `actual` and `expected` here and also add some line returns?

src/helpers.test.js
```diff
@@ -375,6 +378,45 @@ describe('getPrettierEslintOptions', () => {
...
+    };
+    const getUseTabsConfig = () => {
+      const options = getEditorConfigOptions('');
+      return options ? options.useTabs : null;
+    };
+    setIndentStyle('tab');
+    expect(getUseTabsConfig()).toEqual(true);
+    setIndentStyle('space');
+    expect(getUseTabsConfig()).toEqual(false);
+  });
```
robwise: Let's just use two tests here so that it's a bit clearer what's going on and we don't have to write a bunch of custom code. Something like:

```js
  test('returns true if editor config indent_style is set to "tab"', () => {
    editorconfig.parseSync.mockImplementation(() => ({
      indent_style: 'tab',
    }));

    const actual = getEditorConfigOptions('fakeFilepath.js').useTabs;

    expect(actual).toBe(true);
  });
  
  test('returns false if editor config indent_style is set to "spaces"', () => {
    editorconfig.parseSync.mockImplementation(() => ({
      indent_style: 'spaces',
    }));

    const actual = getEditorConfigOptions('fakeFilepath.js').useTabs;

    expect(actual).toBe(false);
  });
```

src/helpers.test.js
```diff
@@ -375,6 +378,45 @@ describe('getPrettierEslintOptions', () => {
   });
 });
 
+describe('getEditorConfigOptions', () => {
+  test('parse editorconfig by filename', () => {
+    getEditorConfigOptions('filename');
+    expect(editorconfig.parseSync).toHaveBeenCalledWith('filename');
+  });
```
robwise: Since we mock the implementation below, I think the above test can be safely eliminated without losing any error checking. We could also just put this expectation in the below test in addition to the other expectation as well.

robwise:
@RoM4iK Arre you going to have a chance to hit these, or should I just merge and then I can do myself (which I'd be happy to do)?

RoM4iK:
@robwise, yes, already started work on this.

robwise:
@RoM4iK Ok awesome!!

RoM4iK:
Hi @robwise, since our `helpers` module became larger, I thought about decomposing an `options` module from it, and did it.

Issue #144, opened 2017-04-27
Make "Format on Save > Enabled" more discoverable


gaearon:
I participated in a workshop aimed at JS beginners yesterday, and Atom Prettier plugin was super helpful to beginners who aren’t feeling comfortable with JS formatting conventions 👍 

One minor piece of feedback is that it’s easy to miss how to enable “Format on Save” in settings, because the checkbox just says “Enabled”. I guess the contrast of section heading might be different depending on the theme and the monitor (e.g. Windows laptops often have very low contrast).

It would help if instead of "Enabled" it said just "Format Files on Save".

robwise:
Awesome to hear! Hmm, I can see what you mean. Yeah, I will change it.

PR #145, opened 2017-04-27
Tweak the checkbox name to be more prominent

gaearon:
Fixes https://github.com/prettier/prettier-atom/issues/144.


robwise:
Thanks!

PR #152, opened 2017-05-03
feat(eslint-fallbacks): fallback to user's prettier settings if cannot be inferred from eslint confi

robwise:
Previously, if we couldn't infer a prettier setting from a user's eslint config, we fell back to
default prettier settings. Now, we take advantage of a new prettier-eslint feature that let's us
provide fallback options to use instead of the prettier defaults. This means that if you don't have
a semicolons rule in your eslint config, we will use the semicolon rule from your prettier-atom
settings.

### DIFF
--- a/dist/config-schema.json
+++ b/dist/config-schema.json
@@ -8,7 +8,7 @@
   },
   "useEslint": {
     "title": "ESlint Integration",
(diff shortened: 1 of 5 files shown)


eriklharper:
What you need to do is update the fallback to do in this order: `eslint` > `.prettierrc` > `prettier-atom`.  Doesn't make much sense to me to have the overrides not look at the `prettier` settings before they default to what the user's Atom settings are set to.  You run into inconsistencies much too easy between developers working on a shared codebase.

robwise:
@eriklharper I'm pretty sure this is how it works, but you'd have to confirm over at the prettier-eslint repo. I'm definitely feeding fallbacks from prettier-atom.

PR #157, opened 2017-05-05
feat(options): add editorconfig option to package settings

RoM4iK:
Add editorconfig option which allows user to enable/disable editorconfig integration

image


robwise:
Awesome, thanks!

PR #158, opened 2017-05-05
Add build.watch task and update yarn.lock

darahak:
I keep forgetting to build my sources, so I wanted to add this.
Also updating `yarn.lock` which was missing dependencies that were added recently.


robwise:
@darahak you need to update the flow typed defs when bump

PR #163, opened 2017-05-10
feat(statusBar): Added click handler to toggle formatOnSave

markacola:
Added a click listener to toggle the formatOnSave option via the status indicator.

### DIFF
--- a/dist/statusTile.js
+++ b/dist/statusTile.js
@@ -3,6 +3,8 @@
 var _require = require('./helpers'),
     getConfigOption = _require.getConfigOption;
 
+var toggleFormatOnSave = require('./toggleFormatOnSave');
+
 var tooltip = null;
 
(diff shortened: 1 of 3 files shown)


robwise:
Thanks! One question, do we need to remove the eventListener upon disposing of the tooltip (to prevent memory bloat)?

markacola:
It's my understanding that the whole element will get garbage collected
when it is removed from the DOM and there is no reference kept around,
including the event listener?
Please correct me if I am wrong.

robwise:
@markacola I think it depends on the environment in which the code is being run as to whether it's smart enough to gc the listener: http://stackoverflow.com/questions/6033821/do-i-need-to-remove-event-listeners

I'm not sure whether Atom is going to do this correctly or not, but I'm leaning towards "yes," so I'm going to merge!

PR #165, opened 2017-05-12
Refactor

robwise:
This is a substantial refactor of the codebase. We had started to amass too many files in the root of the src directory, and the helpers file had become very incohesive. Now, if a certain function is only used once and does not interface with atom or the atom TextEditor object, it stays near the function calling it either in the same file or as a sibling file. Each high-level operation or concern is given its own folder so that the root of src is much cleaner and easier to read.

Other changes of note:
- moved away from the strict `actual`/`expected` syntax for tests as this was unusual for people trying to contribute to the project, and I'm not 100% sold on it personally anyway
- refactored certain functions to take a more FP approach and avoid local state wherever possible
- removed flow coverage of test files: flow seems to have terrible jest support, so we were having to put `$FlowFixMe` annotations everytime we used `mockImplementation`. That noise was probably hurting more than flow was helping.


robwise:
@darahak I have Q/A'd this locally, do you think you could check it out too? The changes are so deep that I'm just worried about it because it's like launching from scratch.

darahak:
@robwise Sure, I'll take a look!

src/executePrettier/getPrettierInstance.js
```diff
@@ -0,0 +1,36 @@
...
+// charypar: This is currently the best way to use local prettier instance.
+// Using the CLI introduces a noticeable delay and there is currently no
+// way to use prettier as a long-running process for formatting files as needed
+//
+// See https://github.com/prettier/prettier/issues/918
+const requireWithFallbackToBundledPrettier = (prettierPackagePath: ?string): typeof bundledPrettier =>
+  // $$FlowFixMe
+  prettierPackagePath ? require(prettierPackagePath) : bundledPrettier; // eslint-disable-line
+
+const getPrettierInstance: (editor: TextEditor) => typeof bundledPrettier = _.flow(
```
darahak: This line is confusing to me because of `(...) => typeof bundledPrettier = ...`.
Is it specific to Flow?

robwise: Yes specific to flow, it's an inline typing of the `getPrettierInstance` `const`. Because calling `_.flow()` returns a function, I don't need to wrap it with another arrow function. However, lodash/fp doesn't have flow-typed support, so I am explicitly defining the API. This says `getPrettierInstance` is a function that takes a `TextEditor` as the only argument and returns something that has the same type as the `bundledPrettier` object.

src/displayDebugInfo/index.js
```diff
@@ -0,0 +1,22 @@
+// @flow
+const readPkg = require('read-pkg');
+const path = require('path');
+const { getAtomVersion, getPrettierAtomConfig, addInfoNotification } = require('../atomInterface');
+
+const getDepPath = (dep: string) => path.join(__dirname, '..', '..', 'node_modules', dep);
+
+const getDebugInfo = () => `
```
robwise: done. fixed using `.trim()`

src/atomInterface/__snapshots__/index.test.js.snap
```diff
@@ -0,0 +1,3 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`getPrettierOptions() returns all prettier options 1`] = `undefined`;
```
robwise: roger, I deleted these snapshots and re-wrote from scratch

decls/index.js
```diff
@@ -36,6 +36,8 @@ declare type Atom$Disposable = any;
 declare type Atom$View = any;
 declare type Atom$Workspace = any;
 declare type Atom$Command = { name: string, displayName: string };
+declare type Atom$Notifications$Options = { detail?: ?string | ?Error, dismissable?: ?boolean };
```
robwise: done.

robwise:
@darahak thanks bro! I fixed the stuff you pointed out. Beta testing went well, just one issue due to a typo so far (which I fixed). If you're curious about learning FP, I find it much easier to read if you do it right once you get used to it. Check out this free guide: https://drboolean.gitbooks.io/mostly-adequate-guide/

He uses ramda but `_.flow` is the same thing as ramda's `composeRight`.

darahak:
@robwise Thanks for the suggestion! I actually used Ramda for one project, so it should be fine.
I agree that it's more readable once you know the concept and what each function does.

PR #185, opened 2017-06-06
Fix Changelog

k15a:


robwise:
hey @k15a, the changelog is actually auto-generated, so I will make a note to fix the description in the commit, thanks!

PR #186, opened 2017-06-06
fix(Settings): Add missing parser options to settings screen

splodingsocks:
Add typescript and postcss as parser options on the settings screen.

I've also added myself to the contributors list. I hope that's okay :).
Thanks!


robwise:
I will merge this for now but I'm going to start working on an implementation where these are in separate sections, that way you don't have to choose between typescript, css, or javascript, but rather can switch between these types of files at will and have it automatically work still.

PR #189, opened 2017-06-08
feat(parsers): add typescript and CSS support

### LINKED ISSUE #183: Support formatting CSS
The latest release includes Prettier 1.4.0, which includes support for CSS formatting, but prettier-atom doesn't seem to be running for CSS files. I tried adding `source.css` to my Scopes list, but that didn't do it.

### LINKED ISSUE #184: Support typescript parser option
Prettier [1.4.0](https://github.com/prettier/prettier/releases/tag/1.4.0) has Typescript support, so i guess we should add it here too?
I am pretty noob in atom extensions, but at first glance just adding it to config-schema.json options will do the thing? If so, i can make PR. Otherwise probably more experienced person should do this :)

robwise:
We now allow adding scopes (some defaults have already been provided for you) for typescript and
CSS. If we recognize the text you are trying to format as being one of these scopes, we will
automatically use the appropriate parser.

Resolves #183 
Resolves #184

### DIFF
--- a/dist/atomInterface/index.js
+++ b/dist/atomInterface/index.js
@@ -1,5 +1,7 @@
 'use strict';
 
+function _toConsumableArray(arr) { if (Array.isArray(arr)) { for (var i = 0, arr2 = Array(arr.length); i < arr.length; i++) { arr2[i] = arr[i]; } return arr2; } else { return Array.from(arr); } }
+
 // constants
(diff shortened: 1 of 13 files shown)


robwise:
I'm not a typescript user and I usually use mostly Sass. I expect there are going to be some bugs with this, but that's okay, we can fix them as they come in. In the meantime, people will hopefully get some usefulness out of these new parsers.

PR #197, opened 2017-06-15
feat(css): Add LESS and SCSS to default scopes for CSS files

### LINKED ISSUE #195: Support *.scss files
`.css` files are now supported in the latest version but `.scss` isn't. It seems like it processes it as javascript:

image

However, when I change the language in the bottom right of atom to `CSS` (instead of CSS) it will format it properly. Am I missing something or is this a bug?

darahak:
Fixes #195

Also made some changes in the package settings: fixed some typos and tried to clarify some descriptions (see comments in #195).

### DIFF
--- a/dist/config-schema.json
+++ b/dist/config-schema.json
@@ -1,21 +1,21 @@
 {
   "silenceErrors": {
     "title": "Silence Errors",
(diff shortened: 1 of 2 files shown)


darahak:
@robwise Thanks for the quick review! Merging.

PR #200, opened 2017-06-16
feat(statusTile): Display status bar tile only if active file matches one of the scopes in settings

### LINKED ISSUE #170: Display status bar tile only if current file matches scope
We should remove the tile if there are no open files or if the language used is not included in our scopes (it doesn't make sense if the user is writing Python for example).
It's cleaner this way, and this is what other packages do (like `linter-eslint`).

darahak:
Fixes #170 

Also updated the tooltip to show the user can click on it to toggle `Format on Save`.

PS: My commit messages were truncated for some reason 😒

### DIFF
--- a/src/statusTile/updateStatusTileScope.js
+++ b/src/statusTile/updateStatusTileScope.js
@@ -1,0 +1,16 @@
+// @flow
+const { getCurrentScope } = require('../editorInterface');
+const { getAllScopes } = require('../atomInterface');
+
+const updateStatusTileScope = (element: HTMLElement, editor: TextEditor) => {
+  if (element) {
+    // The editor can be undefined if there is no active editor (e.g. closed all tabs).
+    // eslint-disable-next-line no-param-reassign
+    element.dataset.prettierMatchScope = editor !== undefined &&
+      getAllScopes().includes(getCurrentScope(editor))
+      ? 'true'
+      : 'false';
+  }
+};
+
+module.exports = updateStatusTileScope;
--- a/src/statusTile/updateStatusTileScope.js
+++ b/src/statusTile/updateStatusTileScope.js
@@ -3,14 +3,14 @@ const { getCurrentScope } = require('../editorInterface');
 const { getAllScopes } = require('../atomInterface');
 
 const updateStatusTileScope = (element: HTMLElement, editor: TextEditor) => {
-  if (element) {
-    // The editor can be undefined if there is no active editor (e.g. closed all tabs).
-    // eslint-disable-next-line no-param-reassign
-    element.dataset.prettierMatchScope = editor !== undefined &&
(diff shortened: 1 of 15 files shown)


src/statusTile/updateStatusTileScope.js
```diff
@@ -0,0 +1,16 @@
+// @flow
+const { getCurrentScope } = require('../editorInterface');
+const { getAllScopes } = require('../atomInterface');
+
+const updateStatusTileScope = (element: HTMLElement, editor: TextEditor) => {
+  if (element) {
```
robwise: prefer guard clauses for exceptional scenarios, this way you don't have to wrap the entire implementation inside of an if statement.

```js
if (!element) return;

// ...
``

File: src/statusTile/updateStatusTileScope.test.js [hunk 2]
```diff
@@ -0,0 +1,31 @@
...
+};
+
+beforeEach(() => {
+  getAllScopes.mockImplementation(() => ['source.js']);
+});
+
+it('sets the match-scope data attribute to true if the editor is in scope', () => {
+  getCurrentScope.mockImplementation(() => 'source.js');
+  const { div } = callUpdateStatusTileScope();
+  expect(div.dataset.prettierMatchScope).toBe('true');
```
robwise: can we separate the "given, when, then" parts of the test with blank lines, just for readability?

darahak:
All done! Thanks @robwise

src/statusTile/updateStatusTileScope.js
```diff
@@ -3,14 +3,14 @@ const { getCurrentScope } = require('../editorInterface');
...
-    element.dataset.prettierMatchScope = editor !== undefined &&
-      getAllScopes().includes(getCurrentScope(editor))
-      ? 'true'
-      : 'false';
-  }
+  if (!element) return;
+
+  // The editor can be undefined if there is no active editor (e.g. closed all tabs).
+  // eslint-disable-next-line no-param-reassign
+  element.dataset.prettierMatchScope = editor !== undefined &&
```
robwise: @darahak What's the practical difference between returning `undefined` and returning `'false'` here?

darahak: We set `display: none` in the `statusTile` stylesheet if this attribute is set to `"false"`, so it would not work if we returned `undefined`.
I'm not sure I understood correctly though, does that answer your question?

robwise: Oh okay, yeah.

I was thinking, if the editor can potentially be undefined here, we need to use a maybe type in the flow annotation (e.g.,`?TextEditor`) to indicate this.

darahak: You're right, I'll change that.

PR #202, opened 2017-06-18
refactor(statusTile): Pass Maybe type for case where TextEditor is undefined

darahak:
Also added a test for this case.

### DIFF
--- a/dist/statusTile/updateStatusTileScope.js
+++ b/dist/statusTile/updateStatusTileScope.js
@@ -7,8 +7,6 @@ var _require2 = require('../atomInterface'),
     getAllScopes = _require2.getAllScopes;
 
 var updateStatusTileScope = function updateStatusTileScope(element, editor) {
-  if (!element) return;
-
   // The editor can be undefined if there is no active editor (e.g. closed all tabs).
   // eslint-disable-next-line no-param-reassign
   element.dataset.prettierMatchScope = editor && getAllScopes().includes(getCurrentScope(editor)) ? 'true' : 'false';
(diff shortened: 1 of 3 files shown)


dist/statusTile/updateStatusTileScope.js
```diff
@@ -11,7 +11,7 @@ var updateStatusTileScope = function updateStatusTileScope(element, editor) {
 
   // The editor can be undefined if there is no active editor (e.g. closed all tabs).
   // eslint-disable-next-line no-param-reassign
-  element.dataset.prettierMatchScope = editor !== undefined && getAllScopes().includes(getCurrentScope(editor)) ? 'true' : 'false';
+  element.dataset.prettierMatchScope = editor && getAllScopes().includes(getCurrentScope(editor)) ? 'true' : 'false';
```
darahak: I had to remove the `!== undefined` test since Maybe types are `null | void`.

src/statusTile/updateStatusTileScope.js
```diff
@@ -2,13 +2,12 @@
 const { getCurrentScope } = require('../editorInterface');
 const { getAllScopes } = require('../atomInterface');
 
-const updateStatusTileScope = (element: HTMLElement, editor: TextEditor) => {
+const updateStatusTileScope = (element: HTMLElement, editor: ?TextEditor) => {
```
robwise: does `HTMLElement` need a `?` also?

darahak: No, it's required. Maybe I should drop the `if (!element) return;`.

robwise: Yeah if it's always there, then no need for the guard clause

darahak: Removed it with last commit.

PR #206, opened 2017-06-27
Correct message for isDisabledIfNotInPackageJson

ericclemmons:
> Does not format on save if `prettier` (or `prettier-eslint`/`prettier-eslint-cli` if using *ESLint integration*) is **not** in your project's `package.json` (dependencies or devDependencies)

Added `not` into the message.  I didn't enable it because I thought this was saying it would run the `prettier` dependency that existed in my `package.json`.


ericclemmons:
Closed instead of having to rebase & all that mess.  👋

PR #246, opened 2017-08-11
Add eslint-plugin-prettier to formatOnSave list 

pbarbiero:
Relying only on `prettier-eslint` and `prettier-eslint-cli` skips a good chunk of applications using prettier, using eslint, but not using one of those manual packages. `eslint-plugin-prettier` should be a good package to also look for to see if prettier is enabled and integrated.


robwise:
@pbarbiero Thanks for the PR! 

1. Can you rename the commit to use the conventional changelog format (the changelog is auto-generated from the commit message).
2. Could you run `build` and commit those changes?

pbarbiero:
Sure

Issue #256, opened 2017-08-28
Support Prettier 1.6 Configurations


k15a:
Prettier 1.6 got released some hours ago and you can now configure Prettier per project. It would be great to support the configuration options with the Atom editor plugin.

In my mind the perfect use case would be:
- use the configuration from the editor plugin to display code in Atom
- use the configuration from the project to save the code

Why?
- we can edit the code with our preferred style
- git diff will show a meaningful diff

If that's not possible it would still be great to support the project specific settings and .prettierignore.

What do you think?

darahak:
> In my mind the perfect use case would be:
> - use the configuration from the editor plugin to display code in Atom
> - use the configuration from the project to save the code

@k15a That's an interesting take, but I don't think it would be a desirable behavior for most people.
Personnally, I would assume the config files prevail no matter what, if they are detected.

darahak:
@robwise FYI there are a few regressions being reported on `1.6.0`.
We can start implementing this, but should wait for these issues to get patched before releasing anything.

robwise:
@darahak Roger that, will probably take me a little to get to this. We need to think about who wins if there is a conflict. Here's an example of what I mean (I bolded the one I think would win in the given scenario):

- prettier-atom sets tabs, **prettier config file sets spaces**
- editorconfig sets tabs, **prettier config file sets spaces**
- **eslint config sets tabs**, prettier config file sets spaces
- **prettier-atom detects that file is json so we use json parser**, prettier config says to use babylon parser

robwise:
I don't know, now that I'm looking into it, I think it doesn't make sense for us to replicate all of the code that prettier already has inside of it when trying to determine how to format a file. We should just defer to the config if it exists.

Here is what I'm proposing we do:

- detect whether or not there is a prettier config (.prettierrc file, prettier.config.js file, or a "prettier" key in your package.json file).
- detect whether or not there is a prettierignore (don't format if there is and the given file is ignored)
- if there is a prettier config, format without passing any options from Atom (prettier will find the config itself and infer its own options from that for us)
- if there is not a prettier config, format and pass the options determined from Atom settings (unless the user chooses a new option we will add like "don't format on save if no prettier config is found")

darahak:
@robwise Sounds good. Not only because we avoid replicating code, but because it's what the users would expect.

robwise:
[...]

PR #261, opened 2017-09-04
Implement Prettier Config

robwise:


MrLoh:
can't wait for this, thanks a lot for the hard work

darahak:
We should consider commiting transpiled files separately. It's a bit difficult to navigate and review big changes.
We can update the contributing guides accordingly.

src/formatOnSave/isPrettierInPackageJson.js
```diff
@@ -1,40 +0,0 @@
-// @flow
```
darahak: I might have misunderstood, but isn't this the feature that lets the package use local Prettier installs instead of the packaged one? I thought it was useful for a lot of devs.

src/executePrettier/executePrettierOnBufferRange.js
```diff
@@ -3,30 +3,32 @@ const _ = require('lodash/fp');
...
-const buildPrettierEslintOptions = require('./buildPrettierEslintOptions');
-const buildPrettierOptions = require('./buildPrettierOptions');
+const { shouldUseEslint, runLinter, getPrettierEslintOptions } = require('../atomInterface');
+const { getCurrentFilePath } = require('../editorInterface');
+const { getPrettierInstance } = require('../helpers');
 const handleError = require('./handleError');
 
 const executePrettier = (editor: TextEditor, text: string) =>
-  getPrettierInstance(editor).format(text, buildPrettierOptions(editor));
+  getPrettierInstance(editor)
```
darahak: I totally missed that point when you described it in https://github.com/prettier/prettier-atom/issues/256

I feel like "use a config file or the defaults" is too radical.
We have no fallback if there is no config, and it reduces our coverage of use cases.

I understand this for ESLint because there are a lot of rules, but it's manageable for Prettier.

Consider this:

- Not all teams or projects work with config files (I'm in this situation).
- The package will not always be used in a project setting: I could hack on a single file without having to define a `.prettierrc` for example.

src/config-schema.json
```diff
@@ -1,218 +1,88 @@
...
-    "order": 1
+    "default": false
   },
-  "useEditorConfig": {
-    "title": "EditorConfig Integration",
+  "scopes": {
+    "order": 2,
+    "title": "Scopes on which Prettier should be run",
     "description":
-      "Use [EditorConfig](http://editorconfig.org/) configuration to load project-based settings for Prettier formatting.<br>Takes precedence over your settings. The following settings will be overriden if they are found in your `.editorconfig`: **Use tabs**, **Tab width**, **Print width**.",
```
darahak: Maybe we should keep this option? At least until https://github.com/prettier/prettier/issues/42 is closed.
It's the third most popular request if you sort issues by 👍 reactions.

robwise: And then do a reverse merge, like if it's in the prettier config's options, use that, if not, use the editor config's options?

src/formatOnSave/isPrettierInPackageJson.js
robwise: This is the one that determines whether or not to run formatOnSave based on whether prettier is installed, if the user has chosen the option. The code that gets the local instance is still in (that's `getPrettierInstance`).

robwise:
@darahak So you think:

 **keep**:
- formatting options 
- editorconfig inference

**remove**:
- blacklist globs
- whitelist globs
- ability to not format on save if prettier is not installed locally
- parser-specific scopes (or maybe we still need this? — will investigate)

src/config-schema.json
darahak: Yes, basically keep it as a fallback for people who don't use config files.

darahak:
@robwise Yes, basically keep formatting options because they are still a good interface for CLI single options, and EditorConfig support because Prettier Config is not a complete replacement IMO (see https://github.com/prettier/prettier/issues/42 for related discussion).

I'm not sure about these two:

> - ability to not format on save if prettier is not installed locally
> - parser-specific scopes (or maybe we still need this? — will investigate)

In the end, it's just my opinion. It would be great to have some feedback from users.
Maybe do a beta or a pre-release?

darahak:
On second thought, would it be a good idea to add an option that says: only use Prettier Config or default options (current proposal)?

If it's disabled, the package is able to fall back on user settings.

darahak:
Current PR for EditorConfig support: https://github.com/prettier/prettier/pull/2760

robwise:
Just an update, still working on this, but I have to wait until Prettier releases `resolveConfig.sync` as otherwise Atom's save doesn't get blocked and we format too late in some cases (also needs the release from prettier-eslint that reverts back to synchronous as well: https://github.com/prettier/prettier-eslint/pull/123)

azz:
Tracking for 1.6.2 which will include .sync() https://github.com/prettier/prettier/issues/2804

robwise:
Okay this is ready for Q/A

PR #272, opened 2017-09-16
feat(error-logging): log errors to Atom's developer tools

robwise:
We were either using the linter package or an in-app popup to display errors, but now we also will
log the error to the console for closer inspection.


darahak:
Good call 👍 
It will be far more convenient, at least for stack traces.

PR #299, opened 2017-10-30
Process postcss files!

jescalan:
Postcss is a popular and common css processor/syntax, this change makes it such that prettier-atom lints postcss files by default :sparkles:

Ref #183


robwise:
@jescalan Thanks! Can you change the commit name to `feat(css-parser-scopes): add postcss` or something so that it will get picked up by our automated changelog?

jescalan:
@robwise sure, done! 👬

robwise:
Thanks!

Issue #105, opened 2017-03-29
Bookmarks are gone after running prettier format


oriSomething:
Bookmarks are gone after running prettier format.
I'm not sure if it's an **atom bug** with files that are updated by changes in the file system. But even if it's an Atom bug. We might need to consider to add a work around it.

![screenshots](https://cloud.githubusercontent.com/assets/2685242/24447159/b67278c0-1478-11e7-8a76-327a50520315.png)

robwise:
@oriSomething Does this happen when prettier-atom is turned off and you turn eslint auto-fix on? If not, then I can look at the linter-eslint repo and try and reverse-engineer how they're working around this.

oriSomething:
Yes, with `linter-eslint` the bookmarks stay in place.

Thanks

robwise:
Okay I will see if we can't reverse-engineer what they're doing

robwise:
@oriSomething I checked and eslint's fix on save _does_ remove bookmarks: 

![](https://cl.ly/110y0F1c0i08/Screen%20Recording%202017-04-02%20at%2005.25%20AM.gif)

I'm not really sure how to get around this because now there's the question of what was I trying to bookmark on that line and where did it go, and that requires a lot more intelligence. If you want to take a crack at it, that would be awesome. Otherwise, I'm not sure how to implement so I'm going to close for now.

oriSomething:
Leave it open. I'll try to address this, though I think the checkbox of this setting will have the title "Experimental"

robwise:
Oh okay, awesome! 

In the meantime, I'm going to cut a release because we just merged #113 which will probably help with removing bookmarks unnecessarily if the text isn't changed (if that was happening).

oriSomething:
After some research and the fact prettier doesn't support `sourcemaps` I think there is no a good solution that can be maid in reasonable amount of time.
For now, there are other "work arounds" or more correct using different ways for bookmarking in Atom:

- [Nuclide outline view](https://nuclide.io/docs/features/outline-view/) - it's not a bookmarks manager but it gives okay way to navigate to symbols location in files unlike atom's <kbd>⌘</kbd>+<kbd>r</kbd>
- [comment-anchors](https://github.com/callodacity/comment-anchors) - which gives you the ability to navigate through `// MARK:` comments as in XCode

Issue #215, opened 2017-06-29
Format undoes all editor folds


kentcdodds:
Here's a gif of the behavior. This happens whether I format manually or hit save to format.

![bad](https://user-images.githubusercontent.com/1500684/27712622-1c088b66-5ce5-11e7-9d51-0f143179b42d.gif)

Not 100% certain you can do anything about this, but maybe? Perhaps keep track of all the folded coded and refold it when you're done?

robwise:
I'm not sure, I'd have to take a deeper look at how code folding is done. Unlike #105, this isn't just based on arbitrary line numbers (which can become inaccurate after formatting), so this one might be more within the realm of possibility.

phil-andrews:
This might be a deal breaker for me unfortunately.

kentcdodds:
Sorry to hear that @phil-andrews. This is an open source project, so if it matters that much to you I'm certain that @robwise would welcome [a pull request](http://makeapullrequest.com/) :smile:

robwise:
I guess he'd rather delete his post instead 😕 

To anyone reading this, I definitely DO appreciate pull requests! If you're new to OSS, check out Kent's free course (see his link). Also, let me know what I can do to make it easier to contribute!

phil-andrews:
The comment I left added nothing to the conversation and shouldn't have been left in the first place so I felt it was better to delete it. I do appreciate the direction towards contributing 👍

kentcdodds:
I'm not sure it's possible to do this and it's not a huge deal to me anyway so I'm going to go ahead and close this :)

PR #310, opened 2017-12-03
perf(executePrettier): Use setTextViaDiff when updating entire buffer

rgreenjr:
Using `setTextViaDiff` instead of `setTextInBufferRange` when formatting the entire buffer improves performance on large files, maintains metadata (such as bookmarks) on lines that haven't changed, and eliminates syntax highlight flickering.

I've never worked with Flow or Jest, so I could use some help implementing a test for this change.

Thanks. 👍


src/executePrettier/executePrettierOnBufferRange.js
```diff
@@ -36,7 +36,13 @@ const executePrettierOnBufferRange = (editor: TextEditor, bufferRange: Range) =>
     return;
   }
 
-  editor.setTextInBufferRange(bufferRange, transformed);
+  const editorBuffer = editor.getBuffer();
+  if (editorBuffer.getRange().isEqual(bufferRange)) {
+    editorBuffer.setTextViaDiff(transformed);
+  } else {
+    editor.setTextInBufferRange(bufferRange, transformed);
+  }
```
robwise: j/c, why this check? Why not always use `setTextViaDiff`? In case the user changed something after we started formatting?

robwise:
I think this should resolve https://github.com/prettier/prettier-atom/issues/105

src/executePrettier/executePrettierOnBufferRange.js
rgreenjr: It's needed to support formatting only selected code when the **Format** command is invoked manually (not on save), since `setTextViaDiff` doesn't take a range parameter.

rgreenjr: Can you help with test case? I've never used Flow or Jest. Here's what I have so far:

```js
it('sets the transformed text via diff when buffer equals entire range of editor', () => {
  editor.getBuffer.mockImplementation(() => ({
    getRange: () => ({ isEqual: () => true }),
    setTextViaDiff: () => bufferRangeFixture,
  }));

  executePrettierOnBufferRange(editor, bufferRangeFixture);

  expect(prettier.format).toHaveBeenCalledWith('const foo = (2);', { useTabs: false });
  // expect(editor.getBuffer.setTextViaDiff).toHaveBeenCalledWith('const foo = 2;');
});
```

But I'm getting the following error:

```sh
sets the transformed text via diff when buffer equals entire range of editor

   expect(jest.fn())[.not].toHaveBeenCalledWith()

   jest.fn() value must be a mock function or spy.
   Received: undefined

     at Object.<anonymous> (src/executePrettier/executePrettierOnBufferRange.test.js:48:43)
         at new Promise (<anonymous>)
         at <anonymous>
     at process._tickCallback (internal/process/next_tick.js:188:7)
```

Any advice? Thanks.

rgreenjr:
Yes, this should resolve #105 and #215. The only caveat is that lines changed during formatting will lose their metadata. But once a file has been saved/formatted, its metadata (bookmarks, folds, etc) will stay in tact through subsequent saves/formats. 😄

src/executePrettier/executePrettierOnBufferRange.js
robwise: > It's needed to support formatting only selected code when the Format command is invoked manually (not on save), since setTextViaDiff doesn't take a range parameter.

Oh, duh! Still, maybe we should put in a var or something just to make it clear why you're doing what you're doing there?

As far as that code snippet, the problem is you're accessing `editor.getBuffer.setTextViaDiff`, but `getBuffer` is a function not an object. Something like this:

```js
  const setTextViaDiffMock = jest.fn();
  editor.getBuffer.mockImplementation(() => ({
    getRange: () => ({ isEqual: () => true }),
    setTextViaDiff: setTextViaDiffMock,
  }));

  executePrettierOnBufferRange(editor, bufferRangeFixture);

  expect(prettier.format).toHaveBeenCalledWith('const foo = (2);', { useTabs: false });
  expect(setTextViaDiffMock).toHaveBeenCalledWith('const foo = 2;');
```

rgreenjr: Your test code worked perfectly – thanks so much. I'll issue a separate PR with it shortly.

rgreenjr: > maybe we should put in a var or something just to make it clear why you're doing what you're doing there?

What if I added a comment that explains the rationale? That would allow a little more exposition.

PR #311, opened 2017-12-03
feat(prettier-stylelint): Add support for prettier-stylelint

### LINKED ISSUE #288: Add support for prettier-stylelint
It would be great if support was added for [prettier-stylelint](https://github.com/hugomrdias/prettier-stylelint). This allows prettier to use a [stylelint](https://stylelint.io/) config for formatting CSS/Sass.

taylon:
fix #288 

This PR adds support for [prettier-stylelint](https://github.com/hugomrdias/prettier-stylelint).

Let me know if there is anything that needs to change =D

### DIFF
--- a/src/config-schema.json
+++ b/src/config-schema.json
@@ -7,6 +7,14 @@
     "default": false,
     "order": 1
   },
+  "useStylelint": {
+    "title": "Stylelint Integration",
+    "description":
+      "Use [prettier-stylelint](https://github.com/hugomrdias/prettier-stylelint) to infer your Prettier settings based on your Stylelint config. If we cannot infer a Prettier setting from your Stylelint config (or if there is none in the current project), we will fallback to using your settings in the *Prettier Options* section.",
+    "type": "boolean",
+    "default": false,
+    "order": 1
+  },
   "useEditorConfig": {
     "title": "EditorConfig Integration",
     "description":
--- a/src/config-schema.json
+++ b/src/config-schema.json
@@ -13,7 +13,7 @@
       "Use [prettier-stylelint](https://github.com/hugomrdias/prettier-stylelint) to infer your Prettier settings based on your Stylelint config. If we cannot infer a Prettier setting from your Stylelint config (or if there is none in the current project), we will fallback to using your settings in the *Prettier Options* section.",
     "type": "boolean",
     "default": false,
-    "order": 1
+    "order": 2
   },
   "useEditorConfig": {
     "title": "EditorConfig Integration",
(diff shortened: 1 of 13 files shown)


src/config-schema.json
```diff
@@ -7,6 +7,14 @@
     "default": false,
     "order": 1
   },
+  "useStylelint": {
+    "title": "Stylelint Integration",
+    "description":
+      "Use [prettier-stylelint](https://github.com/hugomrdias/prettier-stylelint) to infer your Prettier settings based on your Stylelint config. If we cannot infer a Prettier setting from your Stylelint config (or if there is none in the current project), we will fallback to using your settings in the *Prettier Options* section.",
+    "type": "boolean",
+    "default": false,
+    "order": 1
```
robwise: we need to update the orders here

src/executePrettier/executePrettierOnBufferRange.js
```diff
@@ -15,18 +19,30 @@ const executePrettier = (editor: TextEditor, text: string) =>
...
     return error;
   }
 };
 
-const executePrettierOnBufferRange = (editor: TextEditor, bufferRange: Range) => {
+const executePrettierOnBufferRange = async (editor: TextEditor, bufferRange: Range) => {
   const cursorPositionPriorToFormat = editor.getCursorScreenPosition();
   const textToTransform = editor.getTextInBufferRange(bufferRange);
-  const transformed = executePrettierOrPrettierEslint(editor, textToTransform);
+  const transformed = await executePrettierOrIntegration(editor, textToTransform);
```
robwise: Have you verified that with babel-polyfill, Atom will properly wait before continuing on? I used to have a problem using `async` in this repo because Atom wouldn't wait, but I never thought to try using babel-polyfill.

taylon:
Unfortunatelly prettier-stylelint does not have a synchronous version, in fact, stylelint itself don't have it.

The polyfill is only adding a function that is not exported from that file, so it should be all good =D

src/executePrettier/executePrettierOnBufferRange.js
taylon: It will not wait, async/await will still work as Promise so we are making the executePrettierOnBufferRange function asynchronous, which should cause no issues at all, as soon as the promise is resolved runLinter will called normally.

I have been using prettier-atom with this changes for a couple of days and everything has been working normally.

robwise: Roger that!

robwise:
@taylon Can you please rebase when you have a moment, and then I can merge!

taylon:
There we go =D

robwise:
Thanks!

PR #312, opened 2017-12-03
fix(config-schema): add `text.md` to default markdown scopes

### LINKED ISSUE #309: Allow to format on save for markdown files
When the focus is on a markdown file the `Prettier` menu is the status bar is not visible even if the option `Show in status bar` is enabled.

When saving, the js code within fences is not formatted.

When running the command `Prettier:Format` the code is formatted as expected.

pvdlg:
Fix #309


robwise:
@pvdlg Can you please rebase and also make sure to run `yarn run build` so that the `dist` files are built?

pvdlg:
It should be ok now.

robwise:
Thanks!

PR #335, opened 2017-12-31
perf(statusTile): faster array creation and lodash lazy loading

haroldtreen:
### Description

Reduces the activation time for `prettier-atom` from 248ms down to ~60ms.

image

### Changes

- Removes the array spread operator from `atomInterface`. This stops `babel-runtime` from requiring large polyfills (`babel-runtime/helpers/toConsumableArray`). Saves ~70ms.
- Lazy loads `lodash/fp/flow` in `editorInterface`. This function is only used for `getCurrentDir`, but gets loaded whenever `editorInterface` is loaded anywhere. Save ~120ms.


src/editorInterface/index.js
```diff
@@ -10,6 +10,14 @@ const {
...
 } = require('../atomInterface');
 
+let flow;
+const lazyFlow = () => {
+  if (!flow) {
+    flow = require('lodash/fp/flow'); // eslint-disable-line global-require
+  }
+  return flow;
+};
+
```
robwise: Maybe we will be able to revert this in a future PR if we implement babel-plugin-lodash? Or at least we can clean up a bit by implementing that import-lazy lib?

robwise:
@haroldtreen Before I merge, can you run `nps build` to update the `dist` files and amend the commit?

src/editorInterface/index.js
haroldtreen: Ya, I think there's a larger story to be tackled of "What's the pattern for lazy loading?".

This mirrors what's being used in `main.js`, but it can become a bit verbose. 
Another PR for `import-lazy` would be interesting.

haroldtreen:
Wooo! Thanks for the speedy response @robwise 🙌 !

Issue #334, opened 2017-12-31
eslint --fix does not run with ESLint Integration


ozum:
Hi,

Similar to #119, after disabling "Fix errors on save" in `linter-eslint` and enabling ESLint integration, saving doesn't seem to run the `eslint --fix` formatting. 

✓ `eslint --fix` on CLI works as expected.
✓ `prettier-eslint` on CLI works as expected.
✓ `linter-eslint` Atom plugin with "Fix errors on save" works as expected.
✓ `Prettier Atom` plugin works as expected without ESLint integration.
✗ `Prettier Atom` plugin **does not** works as expected with ESLint integration.

For this case, I created a minimal repo: https://github.com/ozum/case-prettier-atom-eslint
Please note that I also included zipped version of .atom folder (**~50 MB**) in the repo.

In that repo `index.js` should be added a dangling comma before save, but it is not added.

Kind Regards,

<details>
<pre>
Atom version: 1.23.1
prettier-atom version: 0.44.0
prettier version: 1.9.2
prettier-eslint version: 8.7.0
prettier-atom configuration: {
  "formatOnSaveOptions": {
    "enabled": true,
    "respectEslintignore": true,
    "showInStatusBar": false,
    "javascriptScopes": [
      "source.js",
      "source.jsx",
      "source.js.jsx",
      "source.babel",
      "source.js-semantic",
      "text.html.basic",
      "text.html.vue"
    ],
    "typescriptScopes": [
      "source.ts",
      "source.tsx",
      "source.ts.tsx"
    ],
    "cssScopes": [
      "source.css",
      "source.less",
      "source.css.less",
      "source.scss",
      "source.css.scss",
      "source.css.postcss"
    ],
    "jsonScopes": [
      "source.json"
    ],
    "graphQlScopes": [
      "source.graphql"
    ],
    "markdownScopes": [
      "source.md",
      "source.gfm",
      "text.md"
    ],
    "excludedGlobs": [],
    "whitelistedGlobs": [],
    "isDisabledIfNotInPackageJson": false,
    "isDisabledIfNoConfigFile": false
  },
  "prettierEslintOptions": {
    "prettierLast": false
  },
  "useEslint": true,
  "useStylelint": false,
  "useEditorConfig": true,
  "prettierOptions": {
    "singleQuote": false,
    "bracketSpacing": true,
    "semi": true,
    "useTabs": false,
    "jsxBracketSameLine": false,
    "printWidth": 80,
    "tabWidth": "auto",
[...]

PR #341, opened 2018-01-04
--arrow-parens is added as a configuration option

### LINKED ISSUE #340: Add configuration option for --arrow-parens
Currently, I cannot find the option in the settings to set value for --arrow-parens parameter.

We extend our Eslint ruleset from Airbnb and arrow-parens rule complains always after the prettier formats the code.

Please add the option to the settings.

ferdibiflator:
fix #340


robwise:
@ferdibiflator thanks for the PR! Could you please amend the commit message so that it will show up in our auto-generated changelog? Something like `feat(options): add arrow-parens prettier option` (or something). Then I can merge!

darahak:
@robwise Not related to this PR, but should the test snapshot look like [this](https://github.com/prettier/prettier-atom/blob/master/src/atomInterface/__snapshots__/index.test.js.snap)? I would expect anything but `undefined`.

ferdibiflator:
@robwise it's ready!

robwise:
@darahak yeah that looks like a mistake, added an issue here: https://github.com/prettier/prettier-atom/issues/342

mauricecruz:
Should there be support for the `"as-needed"` option? We use the airbnb inspired version and this is kinda throwing a wrench in the autofixing
https://eslint.org/docs/rules/arrow-parens#further-reading

robwise:
@mauricecruz we'd need support from prettier to allow doing this, this package is just the bridge between Atom and prettier. In the meantime, if it matters a lot, you could try using the prettier-eslint integration (although there's a bug right now in #334) which will call eslint --fix after running prettier to ensure conformance with your eslint rules.

PR #344, opened 2018-01-07
Exclude non-essential files

vinkla:
This way the users of this package doesn't have to download non-essential files which aren't necessary for the package to function. Such as files used for development (docs, screenshots, tests, .travis.yml, etc).


robwise:
@vinkla this is awesome, can you please amend your commit to use the conventional changelog format? Something like `build(gitattributes): export-ignore unnecessary files` or something?

vinkla:
@robwise done!

robwise:
Thanks!

Issue #322, opened 2017-12-10
Nuclide remote: prettier-atom attempted to set messages with linter package, but was unable. Messages: []


burdakovd:
I'm seeing this error when editing remote file with Nuclide (every time on save I get this error, and can't achieve format-on-save functionality).

```
~/.atom/packages/prettier-atom/dist/linterInterface/index.js:25 prettier-atom attempted to set messages with linter package, but was unable. Messages: []
```

I've applied https://github.com/prettier/prettier-atom/pull/321 locally, and the error still persists.

After adding some more logs around the line where the error is being logged, I see that it was caused by file path being undefined rather than by Linter being null.

The error is logged here: https://github.com/prettier/prettier-atom/blob/e8e8e3fc4da30082a8d70f341b729c46a7771823/src/linterInterface/index.js#L19

In my scenario messages appear to be `[]`, filePath is `undefined`, and delegate is an instance of `IndieLinterDelegate`.

Further, editor from which undefined file path is extracted is a TextEditor that contains remote file in there. Is remote editing something that is supported at all for Prettier?
![image](https://user-images.githubusercontent.com/5193161/33798643-4fd0b96e-dd14-11e7-832e-baccd2cf5a7a.png)

As a workaround I ended up removing `prettier-atom`, integrated `prettier` into `eslint` on the server, and ran `esw --watch --fix` on the server side. Then whenever I'm saving a file from Nuclide, server-side linter attempts to fix it (including formatting according to Prettier), then Nuclide detects that file has  changed remotely and reloads it.

SavePointSam:
I'm located the bug and am working on the fix now.

robwise:
prettier-atom 0.48.1 is released with this fix so I am closing. Please let me know if there continue to be issues, and thanks to @SavePointSam for the fix!

PR #358, opened 2018-01-24
use File path getter over reading directly from property

SavePointSam:
In Nuclide, there is a RemoteFile. This object, while similar to Atom's File, doesn't have the exact same signature. However, they both share the same method `getPath`. Using this method in favor of reading the property directly, repairs prettier-atom's ability to 'format on save' when a user is working with remote files.

Fixes [facebook/nuclide#1443](https://github.com/facebook/nuclide/issues/1443) and [prettier/prettier-atom#322](https://github.com/prettier/prettier-atom/issues/322).

___

Due to an issue with the yarn/npm registry at the moment. I am unable to successfully install dependencies. Various babel packages seem to randomly 404. Therefore, I'm unable to run all the required scripts outlined in CONTRIBUTING.md. What would you like me to do in terms of running tests and adding myself to the contributors list?

I'm also unfamiliar with Flow. I attempted to update the `TextEditor` type definition, but was unsure what needed to happen and opted to avoid modifying it as it exists now. Can someone help me?


src/editorInterface/index.js
```diff
@@ -42,7 +42,8 @@ const isCurrentScopeMarkdownScope = (editor: TextEditor) =>
 
 const isCurrentScopeVueScope = (editor: TextEditor) => getVueScopes().includes(getCurrentScope(editor));
 
-const getCurrentFilePath = (editor: TextEditor) => (editor.buffer.file ? editor.buffer.file.path : undefined);
+const getCurrentFilePath = (editor: TextEditor) =>
+  editor.buffer.file ? editor.buffer.file.getPath() : undefined;
```
darahak: To fix the Flow error on this line (also reported in CI), you need to update the type defs in [decls/index.js](https://github.com/prettier/prettier-atom/blob/master/decls/index.js).

The [documentation](https://atom.io/docs/api/v1.23.3/File#instance-getPath) indicates that `getPath()` returns a `String`, so the `TextEditor` type should be updated like this:

```js
declare type TextEditor = {
  getGrammar: () => { scopeName: string },
  getBuffer: () => { getRange: () => Range, setTextViaDiff: (text: string) => Range },
  getCursorScreenPosition: () => Point,
  getLastCursor: () => { getScopeDescriptor: () => Atom$ScopeDescriptor },
  getSelectedText: () => string,
  getSelectedBufferRanges: () => Ranges,
  getTextInBufferRange: (bufferRange: Range) => string,
  setCursorScreenPosition: (point: Point) => Point,
  setTextInBufferRange: (bufferRange: Range, text: string) => Range,
  buffer: {
    file: ?{
      path: ?FilePath,
      getPath: () => string // HERE
    }
  },
  backwardsScanInBufferRange: (regex: RegExp, Range: Range, iterator: Atom$Iterator) => void,
};
```

darahak:
Thank you for your contribution!

To complete the PR, you will also need to do this:

- [ ] Amend your commit message and follow this template (for the automatic **changelog** updates): `type(scope): message` (see previous [commits](https://github.com/prettier/prettier-atom/commits/master))
- [ ] Run the `build` script and commit the modifications in `/dist`.
- [ ] Run the `addContributor` script if you want to add yourself to the contributor list and commit the changes.

src/editorInterface/index.js
SavePointSam: Thank you!

SavePointSam:
I've gone back and updated my commit message, add the type declaration and was able to add my contributing information. However, I'm unable to run a build as I can't download the dependencies still.

```
~/source/prettier-atom issue-322_nuclide-remote*
❯ yarn
yarn install v1.3.2
[1/4] 🔍  Resolving packages...
[2/4] 🚚  Fetching packages...
error An unexpected error occurred: "https://registry.yarnpkg.com/@babel/types/-/types-7.0.0-beta.31.tgz: Request failed \"404 Not Found\"".
info If you think this is a bug, please open a bug report with the information provided in "/Users/savepointsam/source/prettier-atom/yarn-error.log".
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
```

What would you like to do about this?

SavePointSam:
Alright, I was finally able to install the dependencies. I'll finish up shortly. Apologies for all the commits.

SavePointSam:
Cool, all done!

robwise:
Thanks for the PR!

Issue #34, opened 2017-01-29
Loses cursor position on format


dwwoelfel:
When I trigger a format, my cursor stays on the same line as it was before the code was formatted. It would be nice if the cursor followed its position in the code. 

A quick example: 

The pipe represents my cursor. Before formatting, my cursor is just after `keyTwo` in `mapTwo`. After formatting, I would expect it to move with the code. Instead, it's located just after `valueOne` in `mapOne`.

Before:
```
const mapOne = {keyOne: 'valueOne', keyTwo: 'valueTwo', keyThree: 'valueThree', keyFour: 'valueFour'};
const mapTwo = {keyOne: 'valueOne', keyTwo: |'valueTwo', keyThree: 'valueThree', keyFour: 'valueFour'};

```

After:
```
const mapOne = {
  keyOne: 'valueOne',|
  keyTwo: 'valueTwo',
  keyThree: 'valueThree',
  keyFour: 'valueFour',
};
const mapTwo = {
  keyOne: 'valueOne',
  keyTwo: 'valueTwo',
  keyThree: 'valueThree',
  keyFour: 'valueFour',
};

robwise:
While cool, this isn't really possible to code because it would require some advanced code intelligence and would be too difficult. That being said, if anyone wants to take a crack at it and submit a PR, that would be awesome. Closing for now.

ryanhiebert:
Does prettier/prettier#1637 change things here?

robwise:
@ryanhiebert probably, reopening!

robwise:
See: https://github.com/prettier/prettier#prettierformatwithcursorsource--options

SavePointSam:
Seems pretty neat. I'll take a stab at it.

SavePointSam:
Got a POC up and running. Should have a PR in the next couple days.

SavePointSam:
@robwise, I noticed that you alternate between two different ways of updating the buffer. However the reasoning provided in both the [comment](https://github.com/prettier/prettier-atom/commit/378b6bd7bae1f4e061a6d7ff15ba9497b86779cb#diff-032412b91b2710f2732c7c28fd356332R58) and commit message don't make sense to me given the ability to retain cursor position with this update.

Is this something you'd want to remove to gain the perf benefits? Is there any reason why we wouldn't want to always maintain the cursor position relative to the source given that we have the option now?

[...]

PR #359, opened 2018-01-27
feat(executePrettier): Maintain cursor position on format

SavePointSam:
fixes: prettier-atom #34 

Due to some inconsistencies in how prettier was running tests, I had to slightly modify some tests in order for them to pass and still be useful. I'm looked into what was going on a bit, but for some reason I couldn't determine what was happening. For some reason `formatWithCursor` when running in test mode only returns the formatted text and not in the expected format. However, when running in Atom it works as expected.

I opted to leave the performance issue alone for now as that's a bigger decision than I should make. Though personally I would opt for `setTextViaDiff` by itself. I personally only ever use format on save. There are plenty of other documented solutions to have sub-document formatting done via commit hooks.


robwise:
> Though personally I would opt for setTextViaDiff by itself. I personally only ever use format on save.

Same here, but you know how it is, once you offer a feature it's hard to take it back  ¯\_(ツ)_/¯

src/executePrettier/executePrettierOnBufferRange.js
```diff
@@ -12,60 +11,80 @@ const buildPrettierStylelintOptions = require('./buildPrettierStylelintOptions')
...
+    // TODO: add support for cursor position - https://github.com/hugomrdias/prettier-stylelint/issues/13
+    nextText = await executePrettierStylelint(editor, text);
+  } else if (shouldUseEslint()) {
+    // TODO: add support for cursor position - https://github.com/prettier/prettier-eslint/issues/164
+    nextText = executePrettierEslint(editor, text);
+  } else {
+    return executePrettierWithCursor(editor, text, cursorOffset);
   }
+
+  return { formatted: nextText, cursorOffset: nextCursorOffset };
```
robwise: Let's switch to using returns + guard clauses, I'm not a big fan of `let` although I know this was here before:

```js
const executePrettierOrIntegration = async (editor: TextEditor, text: string, cursorOffset: number) => {
  if (shouldUseStylelint() && isCurrentScopeCssScope(editor)) {
    // TODO: add support for cursor position - https://github.com/hugomrdias/prettier-stylelint/issues/13
    const formatted = await executePrettierStylelint(editor, text);
    return { formatted, cursorOffset };
  } 
  
  if (shouldUseEslint()) {
    // TODO: add support for cursor position - https://github.com/prettier/prettier-eslint/issues/164
    formatted = executePrettierEslint(editor, text);
    return { formatted, cursorOffset };
  } 
  
  return executePrettierWithCursor(editor, text, cursorOffset);
};
```

src/executePrettier/executePrettierOnBufferRange.js
```diff
@@ -12,60 +11,80 @@ const buildPrettierStylelintOptions = require('./buildPrettierStylelintOptions')
...
 const executePrettierOnBufferRange = async (
   editor: TextEditor,
   bufferRange: Range,
   options?: { setTextViaDiff?: boolean },
 ) => {
-  const cursorPositionPriorToFormat = editor.getCursorScreenPosition();
-  const textToTransform = editor.getTextInBufferRange(bufferRange);
-  const transformed = await executePrettierOrIntegration(editor, textToTransform);
+  // grab cursor position and file contents
+  const currentBuffer: TextBuffer = editor.getBuffer();
```
robwise: I was under the assumption that these flow annotations were already auto-inferred, were they not?

src/executePrettier/executePrettierOnBufferRange.js
```diff
@@ -12,60 +11,80 @@ const buildPrettierStylelintOptions = require('./buildPrettierStylelintOptions')
...
+  const cursorPosition: Point = editor.getCursorBufferPosition();
+  const source: string = editor.getTextInBufferRange(bufferRange);
```
robwise: let's keep as `textToTransform`, `source` is a bit nebulous

src/executePrettier/executePrettierOnBufferRange.js
```diff
@@ -12,60 +11,80 @@ const buildPrettierStylelintOptions = require('./buildPrettierStylelintOptions')
...
-  if (_.isError(transformed)) {
-    handleError({ bufferRange, editor, error: transformed });
+  try {
+    results = await executePrettierOrIntegration(editor, source, cursorOffset);
+  } catch (error) {
+    handleError({ editor, bufferRange, error });
     return;
   }
 
+  if (results.formatted === source) return;
```
robwise: let's keep the original logic here so it's super explicit why we return early:

```js
const isTextUnchanged = results.formatted === textToTransform;
if (isTextUnchanged) return;
```

src/executePrettier/executePrettierOnBufferRange.js
```diff
@@ -12,60 +11,80 @@ const buildPrettierStylelintOptions = require('./buildPrettierStylelintOptions')
...
+  const cursorOffset: number = currentBuffer.characterIndexForPosition(cursorPosition);
+  let results: Prettier$CursorResult = {
+    cursorOffset,
+    formatted: source,
+  };
 
-  const isTextUnchanged = transformed === textToTransform;
-  if (!transformed || isTextUnchanged) return;
+  // make sure we actually have text to format
+  if (!source || source.length === 0) return;
```
robwise: prefer variables over comments, although I think if we use lodash it's self-explanatory:

```js
if (_.isEmpty(source)) return;
```

decls/index.js
```diff
@@ -99,6 +107,10 @@ declare type Prettier$SyntaxError = {
   message: string,
   stack: string,
 };
+declare type Prettier$CursorResult = {
+  formatted: any,
+  cursorOffset: number,
+};
```
robwise: if this is just for our own internal use inside of `executePrettierOnBufferRange`, we should declare inside of that file. These typedefs are more for things that are third-party dependencies and/or things that are used globally throughout the app.

SavePointSam: I mean it is the type definition for an external resource. I'm happy to declare in `executePrettierOnBufferRange ` if you still wish for that.

src/executePrettier/executePrettierOnBufferRange.js
SavePointSam: I like this pattern more anyway! I was just trying to adapt based on what was there. Was already having enough fun trying to get flow types figured out that I hadn't taken the time to rethink this here. 👍

src/executePrettier/executePrettierOnBufferRange.js
SavePointSam: I'm not sure, I'm very new to type annotations (this is the first time I've worked with them). Would you suggest removing this one? I figured the more explicit we can be, the better.

src/executePrettier/executePrettierOnBufferRange.js
SavePointSam: I was naming based off the Prettier API documentation. I can switch it back if you prefer. Personally, I prefer to stick with terms that exist in the documentation people would be looking at when working with external systems.

src/executePrettier/executePrettierOnBufferRange.js
SavePointSam: Didn't know that helper existed, wonderful!

src/executePrettier/executePrettierOnBufferRange.js
SavePointSam: Sounds good!

decls/index.js
robwise: Oh sorry, I thought this was just an internal thing you made up. Let's leave for now then 👍

src/executePrettier/executePrettierOnBufferRange.js
robwise: Yeah, let's remove—just as a sort of standardization thing. Flow can be pretty smart about this stuff, but I use the `flow_ide` package and it actually underlines things if flow can't infer them so that I know where it's necessary.

src/executePrettier/executePrettierOnBufferRange.js
robwise: I see both sides on this one, IMO `source` isn't a good name in the context of what's going on here, but yes it is helpful to be consistent with naming.

SavePointSam:
Updated!

robwise:
Awesome job!

SavePointSam:
Probably a good idea. It's a hard one to track down, that's for sure. It only ever appeared when running inside a Jest sandbox.

Issue #361, opened 2018-01-30
prettier-atom failed: Cannot read property 'node' of undefined


dnspnjrs:
Updating to the new `prettier-atom 0.49.0` gives me an error on every save when the last 2 lines are blank.

![screen shot 2018-01-30 at 17 38 37](https://user-images.githubusercontent.com/6238828/35580063-9bd2622a-05e8-11e8-903b-142edbe06daf.png)

How i tried to solve it:
- restarted atom
- reinstalled prettier-atom and its dependencies
- restarted again and again.

```
OSX: version 10.11.6
Atom version: 1.23.3
prettier-atom version: 0.49.0
prettier version: 1.10.2
prettier-eslint version: 8.8.1
prettier-atom configuration: {
  "formatOnSaveOptions": {
    "enabled": true,
    "respectEslintignore": true,
    "showInStatusBar": false,
    "javascriptScopes": [
      "source.js",
      "source.jsx",
      "source.js.jsx",
      "source.babel",
      "source.js-semantic",
      "text.html.basic",
      "text.html.vue"
    ],
    "typescriptScopes": [
      "source.ts",
      "source.tsx",
      "source.ts.tsx"
    ],
    "cssScopes": [
      "source.css",
      "source.less",
      "source.css.less",
      "source.scss",
      "source.css.scss",
      "source.css.postcss"
    ],
    "jsonScopes": [
      "source.json"
    ],
    "graphQlScopes": [
      "source.graphql"
    ],
    "markdownScopes": [
      "source.md",
      "source.gfm",
      "text.md"
    ],
    "vueScopes": [
      "text.html.vue"
    ],
    "excludedGlobs": [],
    "whitelistedGlobs": [],
    "isDisabledIfNotInPackageJson": false,
    "isDisabledIfNoConfigFile": false
  },
  "prettierOptions": {
    "singleQuote": true,
    "bracketSpacing": true,
    "semi": true,
    "useTabs": false,
    "jsxBracketSameLine": false,
    "printWidth": 80,
    "tabWidth": "auto",
    "trailingComma": "none",
    "parser": "babylon",
    "arrowParens": "avoid"
  },
  "useEslint": false,
  "useStylelint": false,
  "useEditorConfig": true,
  "prettierEslintOptions": {
    "prettierLast": false
  }
}
```

<details>
TypeError: Cannot read property 'node' of undefined
[...]

PR #363, opened 2018-02-02
fix(executePrettier): fallback to `format` when `formatWithCursor` fails

SavePointSam:
Turn `formatWithCursor` support into a Progressive Enhancement feature until it stabilizes for all supported parsers.

Works as a temporary fix for #361


robwise:
@SavePointSam can you change commit msg subject from `feat` to `fix` and then I can merge?

SavePointSam:
updated!

robwise:
Thanks! I'll cut a release

Issue #218, opened 2017-07-01
Use prettier-eslint defaults when installed in project


ryanhiebert:
As promised, making my case from my assertion in #216. If the concept is approved, and nobody is too antsy to wait for me to get around to implementing it, I'm willing to put in the work to make it happen.

-------

Right now, using `prettier-eslint` and `eslintConfig` is the only way to configure what `prettier-atom` will do for a project, so it applies to all users of that codebase. Prettier unfortunately loses most of its value if it's not being used consistently across all contributors to a project, unless it's not being enforced by the project at all.

To that end, it's my belief that when the eslint integration is enabled, and is actually installed in the project, the only consistent thing to do would be to use the fallback values defined by `prettier-eslint`. I think that even the eslint integration would ideally be enabled automatically in `prettier-atom` if it's being used by project (noticed by `pretter-eslint` or `prettier-eslint-cli` being installed).

Looking around, I see that there are issues open regarding adding a prettier config file. When that happens, that of course would be a better setting to use as fallbacks, but I'm sure that `prettier-eslint` will use that as it's fallback instead, meaning that the fallback would _still_ not be implemented by `prettier-atom`.

That leaves the settings, ideally, as only being used when there is _no_ other configuration, expressed or implied, given by the project. Anything else makes it difficult to use a shared style through prettier.

--------

In early issues on this front from the prettier repo, the thought was that the editor should reflect the _user's_ preferences, while the saved files should be able to enforce the _project's_ preferences. That's a very interesting idea, but if that's not done, it would be far better to use the project's preferences in the editor, than to default in whole, or especially in part, to the user's preferences.

robwise:
If you are using the eslint integration, and you want everyone in the repo to consistently use, for example, semicolons, you should add that as a rule to your eslint config. That way your linters will catch it in case they forget to run prettier, and everyone's prettier-atom will act exactly the same way. If your prettier-atom is having to fall back to your personal settings, that means you didn't define the rule in your eslint config. I don't see the use case for wanting everyone to use semicolons, and then refusing to add a semicolons rule to your eslint config?

ryanhiebert:
First, thank you for all the hard work that you've put into prettier-atom. It has made my life less mired with dealing with styling, since I know that a quick save will fix things for me, and I can just _move on_ with my life.

-------

[...]

Issue #178, opened 2017-06-02
Show which options are used on format


darahak:
Related to #177.

Looks like `prettier-atom` is not transparent/explicit enough on which options are used for formatting and where they come from (ESLint, EditorConfig, etc.).

It's hard to debug, so we could add another toggle command that shows what the formatter is doing every time.

robwise:
We seem to be getting lots of issues that people think are bugs but are actually just due to not understanding the complex scenarios in which options will change. I think a feature like this could be really useful. Basically, we could make it so that every time we do something, we can optionally log to the console what is happening and why?

darahak:
I agree. At least as a first iteration, console logs would really help.

To take it further, I would limit the output to one message displayed as a notification or a panel (like the `Key Binding Resolver` command).
We could at least avoid having to open DevTools (can be hard to read because of traces from other packages).

Currently, I don't remember how we handle the options. I'll take a look.

olsonpm:
I thought a lot about this today and decided the goal should be to help maintainers debug the resulting issues.  I don't think there's a sensible way to provide a narrow, user friendly amount of data that will let them debug the issues themselves because the apis just don't exist for it.

The best solution I can think of then will be to add all the configurations to the debug output e.g. print the configuration resulting from prettier's `resolveConfig`, eslint's [getConfigForFile](https://eslint.org/docs/developer-guide/nodejs-api#clienginegetconfigforfile) etc.

I'd create a PR with that anyway.
<br>

In order to help users help themselves I believe the following would need to exist
- get the filepaths for each of the configurations found
- get the filepaths for each configuration property of the merged result (merged result = prettier config + eslint config + ...)
- a short explanation of how the configurations were merged.  This would be the responsibility of prettier-atom which understands that e.g. because editorconfig integration is enabled, it should print: "prettier config properties are assigned over those in the editorconfig".

robwise:
This sounds like a good plan. I would only add that it would be preferable to do this via a special debug mode only somehow so as not to spam the console under normal circumstances.

olsonpm:
i was thinking the debug command which new issues advise to copy paste so maintainers have it available.

robwise:
I wouldn't overload that command. Right now it returns a whole bunch of debug output, but it's not necessarily a debug "mode" per se.

olsonpm:
okay, should the new issue template be modified then to request the output of a new command?  Basically if users don't paste it into new issues the i don't think the output provides value

[...]

PR #374, opened 2018-02-08
refactor(buildPrettierOptions.js): Exclusively use local config or pl…

### LINKED ISSUE #370: Request: remove merging of explicit configurations with atom-prettier settings
I just spent some time looking into why my prettier settings in package.json weren't being respected, and it's due to [them being assigned over atom-prettier settings](https://github.com/prettier/prettier-atom/blob/master/src/executePrettier/buildPrettierOptions.js#L70-L74).  I'm wondering what the use-case is for

1. allowing users to `only format if a prettier config is found` and
2. assigning the prettier config over those in atom-prettier.

i.e. why would I want to require an explicit configuration while also using the one local to my editor?  I assume the use-case for allowing users to require both an explicit prettier instance and settings would be to ensure consistent formatting should they move to a different editor or machine.

Even if I set atom-prettier's settings to the defaults, this could become problematic should the defaults in the local prettier instance be different.

Another less breaking solution would be to add an option to ignore atom-prettier's configuration settings.

I'm willing to create a PR if either solution is agreeable.

olsonpm:
…ugin settings

Stop the current behavior of merging a local prettier config with
  those found in the plugin settings.  The reason being a project
  should be formatted according to its own rules, and per Prettier
  convention, omitted rules should fallback to defaults.

Remove the plugin's integration with editorconfig in favor of relying
  on prettier's via its editorconfig option.

BREAKING CHANGE: The formatting configuration is now built according to different rules

fix #370


src/executePrettier/buildPrettierOptions.test.js
```diff
@@ -109,40 +107,46 @@ it('uses vue as the parser if current scope is listed as a Vue SFC scope in sett
...
-  shouldUseEditorConfig.mockImplementation(() => false);
-
   buildPrettierOptions(editor);
+  expect(mockPrettierInstance.resolveConfig.sync.mock.calls[0][1]).toEqual({ editorconfig });
 
-  expect(buildEditorConfigOptions).not.toHaveBeenCalled();
+  editorconfig = true;
+  buildPrettierOptions(editor);
+  expect(mockPrettierInstance.resolveConfig.sync.mock.calls[1][1]).toEqual({ editorconfig });
 });
```
robwise: Can we either:

1. split this into two tests (one for `true` and one for `false`)
2. just test one of the two cases since it shouldn't make a difference whether it's `true` or `false`

(I'd prefer the latter). This way we aren't mutating a variable and doing multiple expectations in a single test.

src/executePrettier/buildPrettierOptions.test.js
```diff
@@ -109,40 +107,46 @@ it('uses vue as the parser if current scope is listed as a Vue SFC scope in sett
...
 
-it('overrides values with editorconfig values if editor config is enabled', () => {
+it('only uses project prettier config when found', () => {
   const editor = buildMockEditor();
   const fakePrettierOptions = { tabWidth: 2, printWidth: 80, useTabs: false };
```
robwise: let's strike this line since it shouldn't be used

src/executePrettier/buildPrettierOptions.test.js
```diff
@@ -109,40 +107,46 @@ it('uses vue as the parser if current scope is listed as a Vue SFC scope in sett
...
-  const fakeEditorConfigOptions = { tabWidth: 4, printWidth: 100, useTabs: true };
-  getPrettierOptions.mockImplementation(() => fakePrettierOptions);
-  buildEditorConfigOptions.mockImplementation(() => fakeEditorConfigOptions);
-  shouldUseEditorConfig.mockImplementation(() => true);
+  const fakePrettierConfigOptions = { tabWidth: 4, printWidth: 100, useTabs: true };
+  const mockPrettierInstance = { resolveConfig: { sync: jest.fn(() => fakePrettierConfigOptions) } };
+  getPrettierInstance.mockImplementation(() => mockPrettierInstance);
   getCurrentFilePath.mockImplementation(() => 'parent-dir/foo.js');
+  getPrettierOptions.mockImplementation(() => fakePrettierOptions);
```
robwise: let's strike this line since it shouldn't be used

src/executePrettier/buildPrettierOptions.test.js
```diff
@@ -109,40 +107,46 @@ it('uses vue as the parser if current scope is listed as a Vue SFC scope in sett
...
+  expect(mockPrettierInstance.resolveConfig.sync.mock.calls[0][1]).toEqual({ editorconfig });
 
-  expect(buildEditorConfigOptions).not.toHaveBeenCalled();
+  editorconfig = true;
+  buildPrettierOptions(editor);
+  expect(mockPrettierInstance.resolveConfig.sync.mock.calls[1][1]).toEqual({ editorconfig });
 });
 
-it('overrides values with editorconfig values if editor config is enabled', () => {
+it('only uses project prettier config when found', () => {
```
robwise: This makes sense in the context of this PR, but in the future, some may not understand what we mean by" only uses project prettier config." I think we should change to `uses settings from the prettier config if found`, and then we can make the test below use `uses settings from prettier-atom if the prettier config is not found`.

src/executePrettier/buildPrettierOptions.test.js
olsonpm: Makes sense.  I appreciate you explaining "why" - I honestly am not a heavy tester so I don't have much knowledge of conventions or best practices.

src/executePrettier/buildPrettierOptions.test.js
olsonpm: good point

src/executePrettier/buildPrettierOptions.test.js
olsonpm: whoops - thanks

robwise:
> I can't think of a scenario where a user would want to use an .editorconfig file with prettier-atom settings.

Although I wouldn't recommend it because of large diff noise, perhaps if a user was working on a project that had not adopted prettier, but had adopted an editorconfig, and the user wants to use prettier for their own personal convenience? I would assume this is the use case that got the editorconfig integration put into the prettier repo in the first place. Because the only other use case is where a project adopts both prettier and an editorconfig but doesn't bother to make sure their prettier config and editor config agree.

So anyway, I feel like there's even more of a use case (if not more) for wanting to respect the editorconfig settings when using personal prettier-atom defaults than when using prettier by itself. Although, I completely overlooked this when first reviewing the PR. So maybe we _should_ keep the editorconfig stuff, but only apply it if a prettier config isn't found?

And the implementation logic would be that we take the editorconfig defaults as the priority (assuming the option to use editorconfig is checked), and then anything that isn't specified, we use the prettier-atom settings.

olsonpm:
> the user wants to use prettier for their own personal convenience

Do you guys want to support that use-case?  Basically allowing someone to format a project file according to settings located on his local machine?  If so then that's totally fine - just not something I agree with.
<br>

> a project adopts both prettier and an editorconfig but doesn't bother to make sure their prettier config and editor config agree.

I see.  I agree it's odd to have a scenario where a prettier configuration overrides an editorconfig setting.  Seems to me this should be presented as an error for the user to fix.
<br>

> maybe we should keep the editorconfig stuff, but only apply it if a prettier config isn't found?

I'll do that.

robwise:
> Do you guys want to support that use-case? Basically allowing someone to format a project file according to settings located on his local machine? If so that's totally fine - just not something I agree with.

This is basically the use case for adding editorconfig to the main prettier repo. Otherwise they could just rely on making sure the prettierrc was set correctly.

olsonpm:
I guess I see it a little differently.  Prettier allows you to fall back to editorconfig defaults instead of prettier defaults.  e.g. this makes sense:

```
// .prettierrc
{
  singleQuote: true
}
```

```
// .editorconfig
[*.js]
indent_style = tab
indent_size = 4
```

It only gets confusing when a prettier config explicitly conflicts with an editorconfig.

olsonpm:
Alright I believe it's working as you intend.  I realized I was unnecessarily building the atom settings object if it wasn't going to be used, so I refactored that into its own function.

I also incorporated a `describe` block to separate the 'config found' and 'config not found' conditions.  Please let me know if the tests should be organized differently.  Also I appreciate your patience.

robwise:
> I guess I see it a little differently. Prettier allows you to fall back to editorconfig defaults instead of prettier defaults

It does allow you to do that sure, but that's not very useful IMO. The case I see much more frequently and I think the reason for them originally implementing this was as a convenience for editor integrations that wanted to support this type of thing. I believe prettier-atom was the only integration that had built its own editorconfig support at the time.

We've gotten lots of prettier-atom devs who either are not wanting to go through the trouble of having to create a config file (or even add prettier locally) or are not allowed to do so by their team (like Kent Dodds), yet still want to use prettier. If these projects do have an editorconfig, then their prettier needs to be compatible with those settings.

src/executePrettier/buildPrettierOptions.js
```diff
@@ -70,8 +72,10 @@ const buildPrettierOptions = (editor: TextEditor) => {
   return {
     ...optionsFromSettings,
     ...buildEditorConfigOptionsIfAppropriate(editor),
-    ...getPrettierConfigOptions(editor),
   };
 };
 
+const buildPrettierOptions = (editor: TextEditor) =>
+  getPrettierConfigOptions(editor) || getOptionsFromSettings(editor);
```
robwise: Does this mean we ignore the user's parser configuration settings if there is a prettier config found? That might be confusing as those settings exist outside the "prettier defaults" section and don't have any indication that they will be ignored.

olsonpm:
ah yeah you're right having the two configs doesn't seem useful.  Not being allowed to add prettier seems weird, but i'm sure it's for good reason.

src/executePrettier/buildPrettierOptions.test.js
```diff
@@ -109,40 +109,60 @@ it('uses vue as the parser if current scope is listed as a Vue SFC scope in sett
...
+    getPrettierInstance.mockImplementation(() => mockPrettierInstance);
+    shouldUseEditorConfig.mockImplementation(() => false);
+    getCurrentFilePath.mockImplementation(() => 'parent-dir/foo.js');
+    getPrettierOptions.mockImplementation(() => fakePrettierOptions);
+
+    const actual = buildPrettierOptions(editor);
+
+    expect(buildEditorConfigOptions).not.toHaveBeenCalled();
+    expect(actual).toEqual(fakePrettierOptions);
+  });
```
robwise: - In the first test, let's add some line returns, so we have a block each for given, when, and then? This allows the reader to jump straight to the setup, or straight to the action under test, or straight to the expectation.
- Usually, when you test you just write the base case and then specify certain edge cases, that way you don't end up with all of your tests having to test that things **didn't** like crazy.
- This "base case" or "happy path" test should be the first test, and then other tests can just specify how they diverge from the path.

All together:

```js
describe('when a prettier config is found', () => {
  it('returns settings from the prettier config', () => {
    const editor = buildMockEditor();
    const fakePrettierConfigOptions = { tabWidth: 4, printWidth: 100, useTabs: true };
    const mockPrettierInstance = { resolveConfig: { sync: jest.fn(() => fakePrettierConfigOptions) } };
    getPrettierInstance.mockImplementation(() => mockPrettierInstance);
    getCurrentFilePath.mockImplementation(() => 'parent-dir/foo.js');

    const actual = buildPrettierOptions(editor);

    expect(actual).toEqual(fakePrettierConfigOptions);
  });
  
  it('tells prettier whether to use editorconfig when resolving options', () => {
    const editor = buildMockEditor();
    const mockPrettierInstance = { resolveConfig: { sync: jest.fn(() => ({})) } };
    getPrettierInstance.mockImplementation(() => mockPrettierInstance);
    shouldUseEditorConfig.mockImplementation(() => true);
    getCurrentFilePath.mockImplementation(() => 'parent-dir/foo.js');
    
    buildPrettierOptions(editor);
    
    expect(mockPrettierInstance.resolveConfig.sync.mock.calls[0][1]).toEqual({ editorconfig: true });
  });
});

describe('when a prettier config is not found', () => {
  it('returns settings from prettier-atom', () => {
    const editor = buildMockEditor();
    const fakePrettierOptions = { singleQuote: true, tabWidth: 2 };
    const mockPrettierInstance = { resolveConfig: { sync: jest.fn(() => null) } };
    getPrettierInstance.mockImplementation(() => mockPrettierInstance);
    shouldUseEditorConfig.mockImplementation(() => false);
    getCurrentFilePath.mockImplementation(() => 'parent-dir/foo.js');
    getPrettierOptions.mockImplementation(() => fakePrettierOptions);

    const actual = buildPrettierOptions(editor);

    expect(actual).toEqual(fakePrettierOptions);
  });

  it('overrides prettier-atom settings with editorconfig values if editor config is enabled', () => {
    const editor = buildMockEditor();
    const fakePrettierOptions = { singleQuote: true, tabWidth: 2 };
    const fakeEditorConfigOptions = { tabWidth: 4 };
    const mockPrettierInstance = { resolveConfig: { sync: jest.fn(() => null) } };
    getPrettierInstance.mockImplementation(() => mockPrettierInstance);
    buildEditorConfigOptions.mockImplementation(() => fakeEditorConfigOptions);
    shouldUseEditorConfig.mockImplementation(() => true);
    getCurrentFilePath.mockImplementation(() => 'parent-dir/foo.js');
    getPrettierOptions.mockImplementation(() => fakePrettierOptions);

    const actual = buildPrettierOptions(editor);

    expect(actual).toEqual({ ...fakePrettierOptions, ...fakeEditorConfigOptions });
  });
});
```

SavePointSam:
A bit of background (as I've been digging through Prettier myself). `.editorconfig` options are overridden by Prettier settings. And `.editorconfig` is only picked up if the `editorconfig` flag is provided when doing configuration resolution. Meaning that you get the following hierarchy:

```
.prettierrc / package.json > .editorconfig > prettier defaults
```

I would suggest that the following pattern be observed for us:

```
.prettierrc / package.json > .editorconfig > prettier-atom settings > prettier defaults
```

It seems [my thoughts on the logic](https://github.com/prettier/prettier-atom/issues/370#issuecomment-362813322) here may have been lost in the shuffle. We are still only using the `prettier-atom` settings in specific use-cases right? (I haven't had time to review the PR, just this conversation.)

src/executePrettier/buildPrettierOptions.js
SavePointSam: True, we only want to discard the prettier configuration settings explicitly, not every setting that `prettier-atom` has. In some advanced cases, a user may want to select a custom parser. We shouldn't take those things away, simply merge over them.

src/executePrettier/buildPrettierOptions.test.js
SavePointSam: We may be able to utilize prettier for the `overrides prettier-atom settings with editorconfig values if editor config is enabled` case. If you ask for settings from prettier with `editorconfig` enabled, it should return an object where it found the editor config and combined those with prettier defaults for us. No need to implement a custom `editorconfig` solution.

SavePointSam:
> So without allowing the user to decide which options should be prettier defaults, we can't accurately decide which ones editorconfig should assign over.

`prettier-atom`'s settings are the Prettier defaults by default, so we don't need to force them to set anything. It's also explicit in the Prettier documentation that the defaults are the final fallback and are opinionated.

> Even if we did allow the user to select defaults, prettier doesn't expose its method of merging prettier options with editorconfig options meaning that code would have to be maintained both in this plugin as well as prettier.

This is exactly what [`prettier#resolveConfig`](https://prettier.io/docs/en/api.html#prettierresolveconfigfilepath-options) does when you provide the `editorconfig` option.

SavePointSam:
> It only gets confusing when a prettier config explicitly conflicts with an editorconfig.

Prettier configs are always the highest priority.

src/executePrettier/buildPrettierOptions.js
SavePointSam: I would suggest moving the Prettier formatting settings into their own block to be ignored if a configuration was already detected and then merge in the parser settings all the time as Prettier should be told which language it is working against, which we know from the editor. :D

robwise:
> It seems my thoughts on the logic here may have been lost in the shuffle. We are still only using the prettier-atom settings in specific use-cases right? (I haven't had time to review the PR, just this conversation.)

@SavePointSam On the contrary, I'm under the impression that this PR is implementing what you were suggesting. Previously we were merging prettierrc and atom settings together, but now we ignore atom settings completely if there is a prettierrc.

src/executePrettier/buildPrettierOptions.test.js
robwise: @SavePointSam but then we won't know what was from the editorconfig (which we want to keep) and which is just a prettier default (which we want to ignore)?

robwise:
@SavePointSam can you give us an example return of `resolveConfig` when `editorconfig` is true? Do the editorconfig settings come in a separate key than the prettier config settings? Or are they merged together? If the latter, we lose the ability to know whether any given setting was determined from the prettierrc or the editorconfig.

SavePointSam:
> can you give us an example return of resolveConfig when editorconfig is true? 

It will merge them into the final payload. They are not isolated in any way. However, we know that only these are supported: `indent_style` `indent_size/tab_width` `max_line_length`. If we need the ability to know which properties came from where, I can write up a proposal to extend the API.

olsonpm:
> prettier-atom's settings are the Prettier defaults by default, so we don't need to force them to set anything.

That's problematic for two reasons
1. As soon as prettier is updated along with its defaults, ambiguity ensues.
2. If we are to stay inline with Prettier's compilation of `prettier config` > `editor config` > `prettier defaults`, and if we assume settings which match prettier defaults should be overwritten by those found in `editor config`, then a user has no way of editing the atom prettier settings to override it.
<br>

> This is exactly what prettier#resolveConfig does when you provide the editorconfig option.
That's not what I meant.  However looking into it, seems prettier now depends on [editorconfig-to-prettier](https://github.com/josephfrazier/editorconfig-to-prettier) which this plugin could do as well instead of [rolling its own](https://github.com/prettier/prettier-atom/blob/master/src/executePrettier/buildEditorConfigOptions.js).  Still I think the ideal scenario would be for prettier to expose a function `prettier.format(string [, prettierConfig [, editorConfig]])` and for the plugin to rely on that instead of merging the options itself, though that's very minor.
<br>

> Prettier configs are always the highest priority.
Functionally yes that's how it works.  But any editorconfig tool that isn't prettier-aware will not be in line with how prettier formats the code and be awkward as a result (I don't use editorconfig, but I imagine it'd be similar to conflicting linter settings where red lines would show on tabs because it doesn't realize prettier is configured to override them to spaces).
<br>

> If we need the ability to know which properties came from where, I can write up a proposal to extend the API.
I think #178 would find that beneficial.

src/executePrettier/buildPrettierOptions.js
olsonpm: Sounds good.  Sorry for overlooking that.

src/executePrettier/buildPrettierOptions.test.js
olsonpm: This is the point I was addressing in my comment on the thread below.  Because there's no way for a user to declare a default in the plugin settings, we can't possibly have a 1-to-1 solution matching prettier's behavior.

olsonpm:
Alright I just pushed an update which fixes the scope-specific settings as well as organizes the tests properly.

For the scope specific settings I left in the [json's `trailingComma`](https://github.com/prettier/prettier-atom/compare/master...olsonpm:use-only-prettier-config-when-found#diff-440e1b1be739422d3e188741e0e71205R53) even though it's not a parser setting.  I assumed this is correct but let me know.

SavePointSam:
> As soon as prettier is updated along with its defaults, ambiguity ensues.

You're not wrong here. We should add a 'use Prettier default' option to each config setting, not assume Prettier defaults in the event they change.

> If we are to stay inline with Prettier's compilation of prettier config > editor config > prettier defaults, and if we assume settings which match prettier defaults should be overwritten by those found in editor config, then a user has no way of editing the atom prettier settings to override it.

This is why `prettier-atom` settings should go after `.editorconfig`, but before actual prettier defaults. Then you don't have this issue as `.editorconfig` has a higher precedent.

> That's not what I meant. However looking into it, seems prettier now depends on editorconfig-to-prettier which this plugin could do as well instead of rolling its own.

Absolutely! Please do this.

> Still I think the ideal scenario would be for prettier to expose a function prettier.format(string [, prettierConfig [, editorConfig]]) and for the plugin to rely on that instead of merging the options itself, though that's very minor.

This puts too much power in the consumers hand, IMO. It will lead to issues that Prettier shouldn't need to support. What happens when an implementation of prettier as an editor plugin has flawed/inconsistent logic of how to merge the options? The Prettier doc say it should do this, but it doesn't. Leaving the configuration resolution in Prettier ensures a cohesive and simple experience for all and prevents various issues that come with more advanced options. It also follows Prettier's stance on being opinionated and normalizes config resolution logic. We have to keep in mind there is a CLI that needs to be supported to, so there will need to be opinionated config resolution regardless of how we consume Prettier.

> But any editorconfig tool that isn't prettier-aware will not be in line with how prettier formats the code and be awkward as a result

This is the end user's responsibility. We can't support conflicting plugins/ideologies.

> I think #178 would find that beneficial.

Sounds good to me! I may take a look once my other WIP stuff is complete. Been rather busy with work and such this past week.

> For the scope specific settings I left in the json's trailingComma even though it's not a parser setting.

Right call! 👍

---

Everything looks good to me. We need to revisit how `buildEditorConfigOptionsIfAppropriate` works, though that falls in line with my current work and is out of the scope of these changes.

Thanks for the work on this!

olsonpm:
> Absolutely! Please do this.

I can tackle this in the current PR if you guys want, otherwise it would make more sense for me to wait for this to be merged and branch from that.
<br>

> This puts too much power in the consumers hand, IMO. It will lead to issues that Prettier shouldn't need to support.

Ah you're right 👍

olsonpm:
One [additional minor change](https://github.com/prettier/prettier-atom/pull/374/commits/46d9f21ac3e0f2e4512f009bdf54ed54b77f5a26#diff-ef0cf7f7d1de7ad6f00d1190065badefL171)

Just removed the incorrect note for the `isDisabledIfNoConfigFile` option:

> *Note:** You must have at least one valid rule in order for your config to be considered present.

SavePointSam:
Let's refactor the .editorconfig parser/reconciler in a seperate PR. :D small scopes!

robwise:
@olsonpm This PR still has WIP in the title, is that still true or okay for me to merge?

robwise:
Also, can we consider this as resolving #218 as well?

olsonpm:
Thanks for checking in with me.  I wanted to thoroughly test it manually before letting it be merged.  I wanted first to make sure an implementation was landed on before I spent a good amount of time manually testing.

That thread is a big one haha, give me a minute.

olsonpm:
Alright it does seem that issue should be resolved with this commit.  I'll confirm though via manual testing.

olsonpm:
Ok - at first I was expecting eslint to follow a similar order to editorconfig e.g. `prettier config` > `eslint config` > `prettier defaults`.  Seems it has its own way of figuring out things especially with the `prettier-atom` setting "Run Prettier Last".  I'm thus assuming the order is fine considering we didn't modify it.

I finished manually testing the following combinations:
- no configs
- .prettierrc
- .editorconfig
- .eslintrc
- .prettierrc and .editorconfig
- .prettierrc and .eslintrc

I didn't test scopes, just a couple settings to prove they were being grabbed/ignored/overridden according to toggles

Anyhoo, this feels ready for me.  If anything goes wrong I'll make sure to work toward fixes as I have plenty of time.

robwise:
@olsonpm SGTM! Awesome job!

PR #376, opened 2018-02-10
refactor(buildEditorConfigOptions.js): Use editorconfig-to-prettier

olsonpm:
- Also inline buildPrettier*Options

BREAKING CHANGE: Derives prettier config options from editorconfig differently


olsonpm:
I was hoping for feedback on how to test buildEditorConfigOptions.js.  A few existing tests broke because they derive the options differently - and I couldn't think of any reasonable solution to simply calling another library.

src/executePrettier/buildEditorConfigOptions.js
```diff
@@ -1,64 +1,11 @@
...
-]);
-
-const mapEditorConfigOptions = (opts?: RawOpts = {}) => ({
-  ...getPrintWidth(opts),
-  ...getTabWidth(opts),
-  ...getUseTabs(opts),
-});
-
-const buildEditorConfigOptions: (file: FilePath) => TranslatedOpts = _.flow(
+const buildEditorConfigOptions: (file: FilePath) => Object = _.flow(
```
robwise: Now that we have so little code in this file, I feel like it's a bit silly. Can we just inline in `buildPrettierOptions`?

Then we could add a test for `buildPrettierOptions` that mocks `editorconfig` and mocks `editorconfig-to-prettier` and asserts that
- editorconfig is called with the correct arguments
- editorconfig-to-prettier is called with the correct arguments
- the results of editorconfig-to-prettier are properly merged in to the final result of `buildPrettierOptions`

olsonpm: i kept it in because buildPrettierEslintOptions had a similarly silly number of lines haha.  I'll make those changes though - thanks

olsonpm:
Alright your changes should be addressed.

Can I inline [buildPrettierStylelintOptions.js](https://github.com/prettier/prettier-atom/blob/master/src/executePrettier/buildPrettierStylelintOptions.js) and [buildPrettierEslintOptions.js](https://github.com/prettier/prettier-atom/blob/master/src/executePrettier/buildPrettierEslintOptions.js) as well?

Also, same as last time I'd still like to test before removing WIP and having it merged.

src/executePrettier/buildPrettierOptions.js
```diff
@@ -10,11 +12,15 @@ const {
   isCurrentScopeVueScope,
 } = require('../editorInterface');
 const { shouldUseEditorConfig, getPrettierOptions, getAtomTabLength } = require('../atomInterface');
-const buildEditorConfigOptions = require('./buildEditorConfigOptions');
 const { getPrettierInstance } = require('../helpers');
 
 const isDefined: (x: any) => boolean = _.negate(_.isNil);
 
+const buildEditorConfigOptions: (file: FilePath) => Object = _.flow(
```
robwise: prefer `{}` over `Object` in general:

The return type used to be this for our internal one, this is probably pretty close:

```js
-type TranslatedOpts = {	
-  tabWidth?: number,	
-  printWidth?: number,	
-  useTabs?: boolean,	
-};	
```

olsonpm: > {} over Object

Ah didn't know about that.  I haven't used flow before so thanks.
<br>

> The return type

Hmm - I purposefully left it at Object (or `{}`) to leave it flexible for changes in the dependency.  Given the yarn.lock file I suppose bumps in version allow us to hardcode its return type.  I'll do this

olsonpm: Which I much prefer to all the requests made by users to authors to support their favorite flavor of js typing haha

olsonpm:
The editorconfigToPrettier type and inlining of `buildPrettier*Options` should be done.

robwise:
LGTM! Let me know when you want to merge

olsonpm:
I'm done testing - I haven't used stylelint before, it's an interesting library.

Unfortunately I found css files don't save like js files do (requires another manual save action).  I believe it's a combination of [this async](https://github.com/prettier/prettier-atom/blob/master/src/executePrettier/executePrettierOnBufferRange.js#L41) and the lack of promise or callback support by [atom's `onWillSave`](https://atom.io/docs/api/v0.187.0/TextBuffer#instance-onWillSave).  Do you think it's worth the bother to send n "add formatSync" PR [prettier-stylelint's](https://github.com/hugomrdias/prettier-stylelint) way?

src/executePrettier/buildPrettierOptions.js
```diff
@@ -10,11 +12,21 @@ const {
   isCurrentScopeVueScope,
 } = require('../editorInterface');
 const { shouldUseEditorConfig, getPrettierOptions, getAtomTabLength } = require('../atomInterface');
-const buildEditorConfigOptions = require('./buildEditorConfigOptions');
 const { getPrettierInstance } = require('../helpers');
 
+type EditorconfigToPretterResult = {
```
SavePointSam: `EditorConfigToPretterResult`

olsonpm: haha I was thinking about that a while when I typed it actually.  I kept finding `const editorconfig = require('editorconfig')` as opposed to `const editorConfig = require('editorconfig');` but I'll change it.

SavePointSam:
LGTM

robwise:
>  Do you think it's worth the bother to send an "add formatSync" PR prettier-stylelint's way?

It already looks like it's synchronous AFAICT?

olsonpm:
[It's not synchronous](https://github.com/hugomrdias/prettier-stylelint/blob/master/src/index.js#L128)
<br>

> Is this a regression with this PR or was always like this?

not a regression.  I didn't git-bisect, just confirmed the behavior on master.

robwise:
Oh it's returning a promise, I just assumed it wasn sync because it didn't have a callback. This is the same as prettier-eslint. All we need to do in this case is add an `await` keyword when invoking, but we can address in a different PR.

PR #378, opened 2018-02-12
fix: Fix asynchronous saves

olsonpm:
Fix an issue where saving css files required another manual save action
  after autoformatting due to stylelint's async api.

- ensure the promise chain stays in tact starting from main.js
  'onWillSave' through executePrettierOnBufferRange.js.


olsonpm:
If you guys prefer more in-depth testing for this then let me know what you're looking for and I'll implement it.

PR #380, opened 2018-02-13
fix(buildPrettierOptions.js): Allow EditorConfig with prettier-atom s…

### LINKED ISSUE #377: Using editor preferences
Updating to `prettier-atom@0.50.0` I've noticed my editor preference for single quotes isn't being respected. It's formatting everything to be double quotes.

I'm only using `prettier`, and have the `eslint` integration disabled. I do have an eslint config, but that also is set to use single quotes.

I'm guessing this may be a result of #374?

```
Atom version: 1.23.3
prettier-atom version: 0.50.0
prettier version: 1.10.2
prettier-eslint version: 8.8.1
prettier-atom configuration: {
  "formatOnSaveOptions": {
    "enabled": true,
    "excludedGlobs": [
      "package.json"
    ],
    "respectEslintignore": true,
    "showInStatusBar": false,
    "javascriptScopes": [
      "source.js",
      "source.jsx",
      "source.js.jsx",
      "source.babel",
      "source.js-semantic",
      "text.html.basic",
      "text.html.vue"
    ],
    "typescriptScopes": [
      "source.ts",
      "source.tsx",
      "source.ts.tsx"
    ],
    "cssScopes": [
      "source.css",
      "source.less",
      "source.css.less",
      "source.scss",
      "source.css.scss",
      "source.css.postcss"
    ],
    "jsonScopes": [
      "source.json"
    ],
    "graphQlScopes": [
      "source.graphql"
    ],
    "markdownScopes": [
      "source.md",
      "source.gfm",
      "text.md"
    ],
    "vueScopes": [
      "text.html.vue"
    ],
    "whitelistedGlobs": [],
    "isDisabledIfNotInPackageJson": false,
    "isDisabledIfNoConfigFile": false
  },
  "prettierOptions": {
    "arrowParens": "always",
    "singleQuote": true,
    "trailingComma": "all",
    "bracketSpacing": true,
    "semi": true,
    "useTabs": false,
    "jsxBracketSameLine": false,
    "printWidth": 80,
    "tabWidth": "auto",
    "parser": "babylon"
  },
  "silenceErrors": false,
  "useEslint": false,
  "useStylelint": false,
  "useEditorConfig": true,
  "prettierEslintOptions": {
    "prettierLast": false
  }
}
```

olsonpm:
…ettings

- Only resolve with editorconfig if a prettier config was found

fix #377

### DIFF
--- a/dist/executePrettier/buildPrettierOptions.js
+++ b/dist/executePrettier/buildPrettierOptions.js
@@ -36,9 +36,23 @@ var isAppropriateToBuildEditorConfigOptions = _.overEvery([isDefined, shouldUseE
 var buildEditorConfigOptionsIfAppropriate = _.flow(getCurrentFilePath, _.cond([[isAppropriateToBuildEditorConfigOptions, buildEditorConfigOptions]]));
 
 var getPrettierConfigOptions = _.cond([[_.flow(getCurrentFilePath, isDefined), function (editor) {
-  return isDefined(getPrettierInstance(editor).resolveConfig.sync) ? getPrettierInstance(editor).resolveConfig.sync(getCurrentFilePath(editor), {
-    editorconfig: shouldUseEditorConfig()
-  }) : null;
+  var hasResolveConfigSync = isDefined(getPrettierInstance(editor).resolveConfig.sync);
+  if (!hasResolveConfigSync) return null;
+
+  var resolveConfigSync = getPrettierInstance(editor).resolveConfig.sync;
+  var filePath = getCurrentFilePath(editor);
+
+  // TODO: when davidtheclark/cosmiconfig#107 is merged, this logic should
+  //   be modified to treat an empty file as an empty object.
+  var prettierConfig = resolveConfigSync(filePath);
+
+  // We only want to resolve with editorconfig when a prettier configuration
+  //   is found in the first place.
+  if (prettierConfig && shouldUseEditorConfig()) {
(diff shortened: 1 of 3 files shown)


robwise:
@olsonpm What's the bug that this one fixes? Is it that if there's no prettier config but there's an editorconfig, it returns a non-null result and we therefore mistakenly think a prettier config exists and fail to fall back to the prettier-atom prettier options? Also, I'm confused as to how the proposed change to cosmic config is able to fix that?

olsonpm:
Currently it's impossible for the `.editorconfig` to integrate with the user's prettier settings configured in the plugin because `resolveConfig(filePath, { editorconfig: true })` returns a configuration.  This is just one of those cases where the API and CLI really tackle two different usecases - personally I think `resolveConfig(filePath, { editorconfig: true })` should return null if no prettier config is found regardless of whether an editorconfig exists.

robwise:
Ok yeah that's what I thought, just wanted to make sure. 

> personally I think resolveConfig(filePath, { editorconfig: true }) should return null if no prettier config is found regardless of whether an editorconfig exists

Yeah I agree. Who would need change in order to do this, Prettier?

olsonpm:
yeah that'd be a prettier change

olsonpm:
thinking about it, i should have modified the commit message to clarify its purpose.  In the future feel free to tell me to change the commit message to something more valuable.

PR #394, opened 2018-03-16
Fix formatOnSave bug with eslintignore

### LINKED ISSUE #347: "Ignore Files in `.eslintignore`" does not work
By default this is enabled.
I added "test" in .eslintignore, so that linting ignores test files. Seems to work. Also disabled prettier formatting for my test files.
I disabled this feature in the prettier-atom config so I can prettify the test files.
I found that it did no prettifying on any of my test files on save.
I restarted atom and found the same.
Lastly, I removed "test" from .eslintignore and, on save of my test file, the prettifying started working again. So it definitely seems like this feature is having no effect.

I've done some research - I'm certain this is a feature specific to prettier-atom, not something in prettier or eslint themselves.

I'm afraid I can't figure out how to run Prettier: Debug (sorry!). But I can tell you I'm on 0.46.0 of the plugin.

olsonpm:
- check "shouldRespectEslintignore" setting prior to checking whether
  the filepath should be eslintignore'd
- add tests for
  : should respect and is eslintignored
  : should not respect and is eslintignored
  : is not eslintignored

fixes #347

### DIFF
--- a/src/formatOnSave/shouldFormatOnSave.js
+++ b/src/formatOnSave/shouldFormatOnSave.js
@@ -9,8 +9,9 @@ const {
   getWhitelistedGlobs,
   isDisabledIfNotInPackageJson,
   isDisabledIfNoConfigFile,
+  shouldRespectEslintignore,
 } = require('../atomInterface');
-const isFilePathEslintignored = require('./isFilePathEslintIgnored');
+const isFilePathEslintIgnored = require('./isFilePathEslintIgnored');
 const isFilePathPrettierIgnored = require('./isFilePathPrettierIgnored');
 const isPrettierInPackageJson = require('./isPrettierInPackageJson');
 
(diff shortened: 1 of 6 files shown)


src/formatOnSave/shouldFormatOnSave.js
```diff
@@ -55,7 +57,7 @@ const shouldFormatOnSave: (editor: TextEditor) => boolean = _.overEvery([
     isFilePathWhitelisted,
     _.overEvery([noWhitelistGlobsPresent, filePathDoesNotMatchBlacklistGlobs]),
   ]),
-  isFilePathNotEslintignored,
+  shouldNotEslintIgnore,
```
robwise: This name doesn't seem right to me? We still format even when we should EslintIgnore (as long as the file doesn't match the eslintignore glob). But this name makes it sound like we are only checking the config option.

What about just inlining the `shouldNotEslintIgnore` function? We do this for the disabling checks below.

olsonpm: I see what you mean.  My intent was to say that we should not ignore it due to eslint.

robwise: Hmm, it seems like the good name for it is already taken: `isFilePathEslintIgnored`. Perhaps we should rename the existing one to something like `doesFilePathMatchEslintIgnore` and then we can use that name?

robwise:
@olsonpm shoot, sorry I didn't catch this before, but do you think you could squash down and use the conventional changelog standard for the commit message? Something like `fix(eslintignore): respect eslintignore setting`. Otherwise, the changelog (which is auto-generated from commit messages) won't pick up the commit.

olsonpm:
ah sorry about that.  Thanks for letting me know

robwise:
Thanks for the fix!

olsonpm:
one of these days I'll stop skimming what people write and actually read it heh

Issue #395, opened 2018-03-16
Plugin support?


kachkaev:
I'm curious if there is anything missing in `prettier-atom` to support [plugins](https://prettier.io/docs/en/plugins.html). After locally installing `prettier` and [`prettier-plugin-elm`](https://github.com/gicentre/prettier-plugin-elm) (which I developed myself 😃 ), the plugin works from the command line, but not in the editor. In particular, pressing `control+option+f` while editing a markdown file, does not format ```` ```elm```` code blocks. Running `yarn prettier same-file.md` from the terminal does this job perfectly.

What could be causing the issue? Can it be that the way prettier is spawned does not let it search for `prettier-plugin-*` in `node_modules`?

<details>
<summary>Prettier: Debug</summary>
<pre>Atom version: 1.24.1
prettier-atom version: 0.52.0
prettier version: 1.11.1
prettier-eslint version: 8.8.1
prettier-atom configuration: {
  "formatOnSaveOptions": {
    "enabled": true,
    "isDisabledIfNoConfigFile": true,
    "showInStatusBar": true,
    "respectEslintignore": true,
    "excludedGlobs": [],
    "whitelistedGlobs": [],
    "isDisabledIfNotInPackageJson": false
  },
  "silenceErrors": true,
  "useEditorConfig": false,
  "useEslint": false,
  "useStylelint": false,
  "prettierOptions": {
    "singleQuote": false,
    "bracketSpacing": true,
    "semi": true,
    "useTabs": false,
    "jsxBracketSameLine": false,
    "printWidth": 80,
    "tabWidth": "auto",
    "trailingComma": "none",
    "parser": "babylon",
    "arrowParens": "avoid"
  },
  "prettierEslintOptions": {
    "prettierLast": false
  },
  "scopes": {
    "javascript": [
      "source.js",
      "source.jsx",
      "source.js.jsx",
      "source.babel",
      "source.js-semantic",
      "text.html.basic",
      "text.html.vue"
    ],
    "typescript": [
      "source.ts",
      "source.tsx",
      "source.ts.tsx"
    ],
    "css": [
      "source.css",
      "source.less",
      "source.css.less",
      "source.scss",
      "source.css.scss",
      "source.css.postcss"
    ],
    "json": [
      "source.json"
    ],
    "graphQl": [
      "source.graphql"
    ],
    "markdown": [
      "source.md",
      "source.gfm",
      "text.md"
    ],
    "vue": [
      "text.html.vue"
    ]
  }
[...]

PR #397, opened 2018-03-22
Fallback to global instance of prettier before falling back to a bundled one

kachkaev:
The PR partially addresses https://github.com/prettier/prettier-atom/issues/395

The extension now searches for a globally installed prettier if a local one is not found; it falls back to a a bundled instance only if both attempts have failed. This change makes it possible to use [Prettier plugins](https://prettier.io/docs/en/plugins.html) when someone does not want to have any npm modules locally. Please note that Prettier still has issues with [detecting plugins in global installs](https://github.com/prettier/prettier/issues/4000), but I hope it gets fixed soon. IMO it's OK to merge this PR before that happens and simply `npm i -g prettier` once the upstream fix arrives.

cc @olsonpm

### DIFF
--- a/src/helpers/getPrettierInstance.js
+++ b/src/helpers/getPrettierInstance.js
@@ -2,13 +2,7 @@
 const _ = require('lodash/fp');
 const bundledPrettier = require('prettier');
 const { getCurrentFilePath } = require('../editorInterface');
-const { findCachedFromFilePath } = require('./general');
-const path = require('path');
-
-const PRETTIER_INDEX_PATH = path.join('node_modules', 'prettier', 'index.js');
-
-const getLocalPrettierPath = (filePath: ?FilePath): ?FilePath =>
-  findCachedFromFilePath(filePath, PRETTIER_INDEX_PATH);
+const { getLocalOrGlobalPrettierPath } = require('./getPrettierPath');
 
 // charypar: This is currently the best way to use local prettier instance.
 // Using the CLI introduces a noticeable delay and there is currently no
@@ -21,7 +15,7 @@ const requireWithFallbackToBundledPrettier = (prettierPackagePath: ?string): typ
 
 const getPrettierInstance: (editor: TextEditor) => typeof bundledPrettier = _.flow(
   getCurrentFilePath,
-  getLocalPrettierPath,
+  getLocalOrGlobalPrettierPath,
   requireWithFallbackToBundledPrettier,
 );
 
--- a/src/helpers/getPrettierInstance.test.js
+++ b/src/helpers/getPrettierInstance.test.js
@@ -5,6 +5,8 @@ const path = require('path');
 const atomLinter = require('atom-linter');
 const createMockTextEditor = require('../../tests/mocks/textEditor');
 const getPrettierInstance = require('./getPrettierInstance');
+const globalModules = require('global-modules');
+const yarnGlobalModules = require('yarn-global-modules')();
 
 test("returns user's project's local prettier instance if it exists", () => {
   const filePath = path.join(__dirname, '..', '..', 'tests', 'fixtures', 'sourceFile.js');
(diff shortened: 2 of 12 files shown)


kachkaev:
I also added printing `git diff` to the CI script, which should run when dirty files are found. Had to re-commit four times because `checkForDistChanges` [was failing](https://travis-ci.org/prettier/prettier-atom/pull_requests) and I could not figure out why. Tried not to commit dist and even tweaked something there manually before simply realising that the problem was with `yarn.lock` not being up-to-date. I did not notice `yarn.lock` initially and thought you were using `npm`. Not a surprise that a new package in `package.json` broke `yarn.lock` and was making this file dirty in CI.

src/helpers/getPrettierInstance.test.js
```diff
@@ -22,6 +23,24 @@ test("returns user's project's local prettier instance if it exists", () => {
   );
 });
 
+test("returns global prettier if user's project has no local prettier package", () => {
+  atomLinter.findCached.mockImplementation(() => undefined);
```
robwise: this seems to be dead code as you're overriding your mock on line 32?

src/helpers/getPrettierInstance.test.js
```diff
@@ -22,6 +23,24 @@ test("returns user's project's local prettier instance if it exists", () => {
...
+  const filePath = path.join(__dirname, 'sourceFile.js');
+  const file = { path: filePath, getPath: () => filePath };
```
robwise: I believe we only need `getPath` to be defined? Well, doesn't matter I guess

src/helpers/getPrettierInstance.js
```diff
@@ -4,11 +4,13 @@ const bundledPrettier = require('prettier');
 const { getCurrentFilePath } = require('../editorInterface');
 const { findCachedFromFilePath } = require('./general');
 const path = require('path');
+const { findCached } = require('atom-linter');
+const globalModules = require('global-modules');
```
robwise: Does this work for globally installed packages in Yarn? The repo makes it sound like it only works for npm? We should really support both if possible.

src/helpers/getPrettierInstance.test.js
```diff
@@ -22,6 +23,24 @@ test("returns user's project's local prettier instance if it exists", () => {
...
+  const editor = createMockTextEditor({ buffer: { file } });
+  const prettierLib = path.join(__dirname, '..', '..', 'tests', 'fixtures', 'prettier.js');
```
robwise: I'm worried that this is a bit confusing because you're using a local prettier instance as a fake prettier instance to use when there is no local prettier instance (it's even confusing to say in English). I think if just renamed it a bit it would be more clear, something like `fakeGloballyInstalledPrettier`?

olsonpm:
mind adding a line in `displayDebugInfo/index.js` so we have this available when helping users with bugs?

kachkaev:
@olsonpm which line?

olsonpm:
just to be clear, I was just hoping for the version of the user's global prettier if found.  If not, I would like that to be explicit as well e.g. "global prettier version: <not found>"

kachkaev:
@olsonpm I'm printing prettier's path in debug info now. Could you please help me with passing editor and thus `filePath` in there?

olsonpm:
If you have prettier's path, then wouldn't you just

```js
const packageJsonPath = path.join(pathToGlobalPrettier, 'package.json')
const globalPrettierVersion = require(packageJsonPath).version
```

Not sure what you're asking

kachkaev:
Got you. Now finally going to sleep 😅

kachkaev:
@robwise  if the PR looks OK to you now after refactoring, it'd really help if you could merge it and release a new version. Can't wait to start using [prettier-plugin-elm](https://www.npmjs.com/package/prettier-plugin-elm) with colleagues! 😄

robwise:
LGTM!

krugar:
This does not find a globally installed `prettier` under linux when node is installed via [nvm](https://github.com/creationix/nvm). Might be related to jonschlinkert/global-prefix#19, but I didn't really care to dig, as this does not hamper use of `prettier-atom` in any way (it falls back gracefully to the bundled `prettier`).

I just wanted to give you folks a heads up.

robwise:
Thanks for reporting, I will make an issue anyway just in case someone is looking for this info: #448

Issue #98, opened 2017-03-22
Disabling prettier in specific projects?


IljaDaderko:
Hi,

I've started using prettier in my personal projects and it is really helping out with code formatting. I am using atom package and have following problem.

As it is my main editor I also use it for my work where we haven't currently introduced prettier, as I have it configured to format on save it causes lint errors in work project, hence I need to enable and disable the package when switching between personal and work projects.

I wanted to check if there is a way to disable prettier's atom package automatically in specific projects somehow?

robwise:
@IljaDaderko Currently your options are limited to using the whitelist globs or blacklist globs. So, the idea being you put matching globs for your projects in there.

However, we are currently working on getting the ability to choose not to format with prettier/prettier-eslint if we can't find it in your package.json (https://github.com/prettier/prettier-atom/issues/59), so that may be an attractive option for you in the future.

dsingleton:
Hi @robwise,

I'm in a similar situation as @IljaDaderko - i'd like to try prettier on a subset of projects. 

Something along the lines of the [`standard-formatter` atom plugin](https://github.com/stephenkubovic/atom-standard-formatter) would be ideal, which is what I think you were describing originally?

Did this get solved in the end? The docs indicate `prettier` formatting is always applied, but I might be misreading it.

> What Version of Prettier Gets used?
> By default, we use the prettier instance in your project's 'node_modules' directory. If one isn't found, then we fall back to using the version that comes bundled with the prettier-atom package.

Thanks!

olsonpm:
we currently have an option to only format on save if you have prettier added as a dependency, or if you have a prettier configuration.  Do these work for you?  I know it's not exactly what you're looking for but it may suffice

dsingleton:
@olsonpm Perfect! Exactly what I was looking for. And thank you for the quick response 😄 

I hadn't installed the package and was going by the docs along. Would you be open to a PR tweaking the docs to mention that option?

olsonpm:
Maybe a small preface under the Detailed Usage header, something like:

*Note: only a few prettier-atom settings are explained below.  You may be interested in exploring all of them in the plugin settings page*

What do you think?

We didn't want to duplicate all the settings documentation, and those detailed in the readme subjectively cover some common use-cases.

PR #399, opened 2018-03-26
docs: Note on settings in readme and fix addContributor snippet

dsingleton:
Hi 👋  First contribution to this project, so apologies if i've missed a part of the process.

In https://github.com/prettier/prettier-atom/issues/98#issuecomment-375997685 @olsonpm and I discussed adding a line to the README to explain it didn't cover all the settings (which had led me to assume a setting I wanted didn't exist).

In the process I added myself to the contributors list and found (what I think is) a bug/typo in the instructions for adding a contributor.


.all-contributorsrc
```diff
@@ -284,6 +284,16 @@
...
+      "name": "David Singleton",
+      "avatar_url": "https://avatars2.githubusercontent.com/u/63201?v=4",
+      "profile": "http://dsingleton.co.uk",
+      "contributions": [
+        "doc"
+      ]
     }
-  ]
+  ],
+  "repoType": "github"
```
dsingleton: I was confused by this bit of the diff, but I think it could be because [`all-contributors-cli` was upgraded](https://github.com/prettier/prettier-atom/commit/0d62f94d17ea8ae8d5329e4a8f57aca01b40ec92) since the last contributor was added.

robwise: Ah yes that's probably what it is, no worries

robwise:
Thanks for the PR!

PR #400, opened 2018-03-27
Follow the UI theme's green color

varemenos:
Currently prettier-atom is using a hardcoded green color that doesn't follow the success(green) color variable that's documented in Atom's UI styleguide (which is set but the currently used UI theme).

There are 2 ways to use this color, passed as prop via the Less variable or via the `text-success` class.

You should also utilise the Octicons provided from the styleguide instead of the hardcoded `content: ' ✓';` value which you can find in the same file.

_you can view the Styleguidw page by typing Ctrl + Shift + Cmd + G_.

### DIFF
--- a/styles/statusTile.less
+++ b/styles/statusTile.less
@@ -12,7 +12,7 @@ div {
 
   // If Format on Save is enabled, make the label more visible.
   &[data-prettier-format-on-save='enabled'] {
-    color: #8bc34a;
+    color: @text-color-success;
     &::after {
       content: ' ✓';
     }


varemenos:
No that's alright, the PR is very small anyway.
Really appreciate the offer though, thanks.

robwise:
@SavePointSam please don't merge PRs when the commit messages aren't formatted using conventional changelog.

SavePointSam:
oh right! Sorry about that

PR #404, opened 2018-03-30
Simplify Settings by Relying on Users Setting Prettier Config

### LINKED ISSUE #178: Show which options are used on format

### LINKED ISSUE #395: Plugin support?

### LINKED ISSUE #398: Tree sitter and default config
The latest version of Atom brought experimental tree sitter support. When enabled, prettier format on save doesn't work. It looks like the grammar names have changed within tree sitter.

I can of course override this & add in the new languages. However when tree sitter eventually becomes default prettier will no longer work out the box.

<details>
Atom version: 1.25.0
prettier-atom version: 0.52.0
prettier version: 1.11.1
prettier-eslint version: 8.8.1
prettier-atom configuration: {
  "formatOnSaveOptions": {
    "enabled": true,
    "isDisabledIfNoConfigFile": true,
    "respectEslintignore": true,
    "showInStatusBar": false,
    "excludedGlobs": [],
    "whitelistedGlobs": [],
    "isDisabledIfNotInPackageJson": false
  },
  "prettierEslintOptions": {
    "prettierLast": true
  },
  "prettierOptions": {
    "bracketSpacing": false,
    "singleQuote": true,
    "trailingComma": "all",
    "semi": true,
    "useTabs": false,
    "jsxBracketSameLine": false,
    "printWidth": 80,
    "tabWidth": "auto",
    "parser": "babylon",
    "arrowParens": "avoid"
  },
  "useEslint": false,
  "useStylelint": false,
  "useEditorConfig": true,
  "scopes": {
    "javascript": [
      "source.js",
      "source.jsx",
      "source.js.jsx",
      "source.babel",
      "source.js-semantic",
      "text.html.basic",
      "text.html.vue"
    ],
    "typescript": [
      "source.ts",
      "source.tsx",
      "source.ts.tsx"
    ],
    "css": [
      "source.css",
      "source.less",
      "source.css.less",
      "source.scss",
      "source.css.scss",
      "source.css.postcss"
    ],
    "json": [
      "source.json"
    ],
    "graphQl": [
      "source.graphql"
    ],
    "markdown": [
      "source.md",
      "source.gfm",
      "text.md"
    ],
    "vue": [
      "text.html.vue"
    ]
  }
}
</details>

### LINKED ISSUE #401: Support proseWrap option
Currently, there is no way to configure the [Prose Wrap option](https://prettier.io/docs/en/options.html#prose-wrap) alongside all of the other Prettier options in prettier-atom's config.

Is this an intentional omission?

### LINKED ISSUE #406: "Cannot read property 'sync' of undefined" on projects without Prettier config
I'm receiving the following error every time I save a file in projects which **don't** have a Prettier config file. I have to disable "Format On Save" to avoid error notifications popping up every time I save a file.

error

I don't have this issue when editing files in projects which contain a Prettier config file.

### LINKED ISSUE #433: TypeScript Not Formatted on Save
I'm not quite sure where to begin on providing triage up front for this one. I have a mixed project, where both ESLint (latest) and TSLint (latest) exist. Prettier is installed (latest) and works great with on the JavaScript files in the project, formats as expected on save. However, Prettier seemingly does nothing on the TypeScript files in the project. No errors, no warnings, no formatting on save.

I searched the issues here and couldn't find anything related. I did find one that had to do with out of date dependencies, but I'm up to date. I'm a bit baffled.

`Prettier:Debug` Output:
```
Atom version: 1.28.2
prettier-atom version: 0.54.0
prettier: bundled
prettier version: 1.13.4
prettier-eslint version: 8.8.1
prettier-atom configuration: {
  "useEslint": false,
  "useStylelint": false,
  "useEditorConfig": true,
  "formatOnSaveOptions": {
    "enabled": false,
    "respectEslintignore": true,
    "showInStatusBar": false,
    "excludedGlobs": [],
    "whitelistedGlobs": [],
    "isDisabledIfNotInPackageJson": false,
    "isDisabledIfNoConfigFile": false
  },
  "prettierOptions": {
    "singleQuote": false,
    "bracketSpacing": true,
    "semi": true,
    "useTabs": false,
    "jsxBracketSameLine": false,
    "printWidth": 80,
    "tabWidth": "auto",
    "trailingComma": "none",
    "parser": "babylon",
    "arrowParens": "avoid"
  },
  "prettierEslintOptions": {
    "prettierLast": false
  },
  "scopes": {
    "javascript": [
      "source.js",
      "source.jsx",
      "source.js.jsx",
      "source.babel",
      "source.js-semantic",
      "text.html.basic",
      "text.html.vue"
    ],
    "typescript": [
      "source.ts",
      "source.tsx",
      "source.ts.tsx"
    ],
    "css": [
      "source.css",
      "source.less",
      "source.css.less",
      "source.scss",
      "source.css.scss",
      "source.css.postcss"
    ],
    "json": [
      "source.json"
    ],
    "graphQl": [
      "source.graphql"
    ],
    "markdown": [
      "source.md",
      "source.gfm",
      "text.md"
    ],
    "vue": [
      "text.html.vue"
    ]
  }
}
```

robwise:
Resolves #395 (plugins will be free to run since settings are no inferred from the prettier config)
Resolves #401 (prose option will just be inferred from prettier config instead of needing to be set in prettier-atom)
Resolves #398 (prettier should just work for experimental features since prettier-atom no longer needs to duplicate new experimental options)
Resolves #178 (prettier-atom doesn't pass any options so there's no longer a need to clarify this)
Resolves #433 (prettier-atom will no longer rely on scopes to determine whether to format on save, it will ask Prettier to figure it out)
Resolves #406 (prettier-atom will now check Prettier version using a slightly different method so this logic bug is no longer a problem)


robwise:
I still need to do some manual Q/A as well as check out whether we can get rid of the new `scopes` option I added (which simply determines whether or not prettier-atom should run based on the current Atom scope, as opposed to the multiple scope options we had previously that did this as well as chose which parser to use. Now we just let prettier decide based on the filepath).

robwise:
Since this is going to break all builds that do not use the very latest prettier, we might as well refactor to use this as well: https://github.com/prettier/prettier/pull/3846

kachkaev:
@robwise WDYT of using Prettier via CLI rather than through `require()`? This way Atom would resolve and use the most up-to-date instance at all times. Otherwise, upgrading Prettier or, say, installing a global instance does not change the way formatting works until Atom is relaunched. I guess this is to do with how required modules are cached by Electron and it would be great if this side-effect was avoided. CLI approach should also make it easier to discover the global instance because the extension can simply try `exec('prettier', [...])` and see if it works at all or not.

kachkaev:
> we're caching this ourselves in the package on purpose to reduce time to format, it's not an Electron thing
Does this mean that we can cache multiple projects' Prettier in paraller? That seems a pretty important thing to have.

Imagine I have a single Atom window with three projects: one using local Prettier 1.13.1 with no plugins, another using Prettier 1.14.0 with a couple of local plugins and a third one falling back to a global Prettier, which has a globally installed plugin (different from project two). I can imagine how this can work when Prettier is called via CLI, but how about JS API?

robwise:
@kachkaev We're using `findCached` from `atom-linter` to do this. You'd have to dive into the internals of that, but I believe It's all filename-based so the Atom project shouldn't be relevant AFAIK.

robwise:
@kachkaev This should be working now. Do you want to test it for a bit (I will do the same), especially with plugins enabled?

kachkaev:
Many thanks @robwise!

I’m on holidays now, so won’t be able to test things in the next week. But I trust your experience!

robwise:
Does everyone think it would be okay to release this finally? I'm super scared because this might break a lot of projects for a lot of people.

kachkaev:
I guess this should happen sooner or later anyway given the direction into which Prettier goes. If you anticipate a lot of broken workflows, how about adding one paragraph to README that would suggest how to temporary roll back?

This command should be a remedy while people migrate from eslint integration etc.:

```
apm install prettier-atom@v0.54.0
```

despairblue:
I just tried this PR and I get multiple errors:

<details>
  <summary>
  errors
  </summary

/home/despairblue/github/prettier-atom/dist/atomInterface/index.js:59 TypeError: Cannot read property 'sync' of undefined
at getPrettierFileInfoForCurrentFilePath (/home/despairblue/github/prettier-atom/dist/helpers/isFileFormattable.js:9)
at /home/despairblue/github/prettier-atom/node*modules/lodash/lodash.min.js:49
at n (/home/despairblue/github/prettier-atom/node_modules/lodash/lodash.min.js:5)
at /home/despairblue/github/prettier-atom/node_modules/lodash/lodash.min.js:51
at u (/home/despairblue/github/prettier-atom/node_modules/lodash/lodash.min.js:5)
at /home/despairblue/github/prettier-atom/node_modules/lodash/lodash.min.js:51
at n (/home/despairblue/github/prettier-atom/node_modules/lodash/lodash.min.js:5)
at /home/despairblue/github/prettier-atom/node_modules/lodash/lodash.min.js:63
at n (/home/despairblue/github/prettier-atom/node_modules/lodash/lodash.min.js:5)
at /home/despairblue/github/prettier-atom/node_modules/lodash/lodash.min.js:51
at u (/home/despairblue/github/prettier-atom/node_modules/lodash/lodash.min.js:5)
at /home/despairblue/github/prettier-atom/node_modules/lodash/lodash.min.js:51
at n (/home/despairblue/github/prettier-atom/node_modules/lodash/lodash.min.js:5)
at /home/despairblue/github/prettier-atom/node_modules/lodash/lodash.min.js:63
at n (/home/despairblue/github/prettier-atom/node_modules/lodash/lodash.min.js:5)
at /home/despairblue/github/prettier-atom/node_modules/lodash/lodash.min.js:97
at n (/home/despairblue/github/prettier-atom/node_modules/lodash/lodash.min.js:5)
at /home/despairblue/github/prettier-atom/node_modules/lodash/lodash.min.js:63
at /home/despairblue/github/prettier-atom/node_modules/lodash/lodash.min.js:49
at attemptWithErrorNotification (/home/despairblue/github/prettier-atom/dist/formatOnSave/index.js:14)
at /home/despairblue/github/prettier-atom/dist/atomInterface/index.js:57
at Generator.next (<anonymous>)
at step (/home/despairblue/github/prettier-atom/node_modules/babel-runtime/helpers/asyncToGenerator.js:17)
at /home/despairblue/github/prettier-atom/node_modules/babel-runtime/helpers/asyncToGenerator.js:35
at new Promise (<anonymous>)
at new F (/home/despairblue/github/prettier-atom/node_modules/core-js/library/modules/\_export.js:36)
at /home/despairblue/github/prettier-atom/node_modules/babel-runtime/helpers/asyncToGenerator.js:14
at attemptWithErrorNotification (/home/despairblue/github/prettier-atom/dist/atomInterface/index.js:65)
at safeFormatOnSaveIfAppropriate (/home/despairblue/github/prettier-atom/dist/formatOnSave/index.js:14)
at /home/despairblue/github/prettier-atom/dist/main.js:38
at Generator.next (<anonymous>)
at step (/home/despairblue/github/prettier-atom/node_modules/babel-runtime/helpers/asyncToGenerator.js:17)
at /home/despairblue/github/prettier-atom/node_modules/babel-runtime/helpers/asyncToGenerator.js:35
at new Promise (<anonymous>)
at new F (/home/despairblue/github/prettier-atom/node_modules/core-js/library/modules/\_export.js:36)
at /home/despairblue/github/prettier-atom/node_modules/babel-runtime/helpers/asyncToGenerator.js:14
at lazyFormatOnSave (/home/despairblue/github/prettier-atom/dist/main.js:42)
at subscriptions.add.editor.getBuffer.onWillSave (/home/despairblue/github/prettier-atom/dist/main.js:111)
at Function.module.exports.Emitter.simpleDispatch (/usr/share/atom/resources/app.asar/node_modules/event-kit/lib/emitter.js:25)
at Emitter.<anonymous> (/usr/share/atom/resources/app.asar/node_modules/event-kit/lib/emitter.js:154)
at Emitter.module.exports.Emitter.emitAsync (/usr/share/atom/resources/app.asar/node_modules/event-kit/lib/emitter.js:157)
at /usr/share/atom/resources/app.asar/node_modules/text-buffer/lib/text-buffer.js:1329
(anonymous) @ /home/despairblue/github/prettier-atom/dist/atomInterface/index.js:59
step @ /home/despairblue/github/prettier-atom/node_modules/babel-runtime/helpers/asyncToGenerator.js:17
(anonymous) @ /home/despairblue/github/prettier-atom/node_modules/babel-runtime/helpers/asyncToGenerator.js:35
F @ /home/despairblue/github/prettier-atom/node_modules/core-js/library/modules/\_export.js:36
(anonymous) @ /home/despairblue/github/prettier-atom/node_modules/babel-runtime/helpers/asyncToGenerator.js:14
attemptWithErrorNotification @ /home/despairblue/github/prettier-atom/dist/atomInterface/index.js:65
safeFormatOnSaveIfAppropriate @ /home/despairblue/github/prettier-atom/dist/formatOnSave/index.js:14
(anonymous) @ /home/despairblue/github/prettier-atom/dist/main.js:38
step @ /home/despairblue/github/prettier-atom/node_modules/babel-runtime/helpers/asyncToGenerator.js:17
(anonymous) @ /home/despairblue/github/prettier-atom/node_modules/babel-runtime/helpers/asyncToGenerator.js:35
F @ /home/despairblue/github/prettier-atom/node_modules/core-js/library/modules/\_export.js:36
(anonymous) @ /home/despairblue/github/prettier-atom/node_modules/babel-runtime/helpers/asyncToGenerator.js:14
lazyFormatOnSave @ /home/despairblue/github/prettier-atom/dist/main.js:42
subscriptions.add.editor.getBuffer.onWillSave @ /home/despairblue/github/prettier-atom/dist/main.js:111
module.exports.Emitter.simpleDispatch @ /usr/share/atom/resources/app.asar/node_modules/event-kit/lib/emitter.js:25
(anonymous) @ /usr/share/atom/resources/app.asar/node_modules/event-kit/lib/emitter.js:154
module.exports.Emitter.emitAsync @ /usr/share/atom/resources/app.asar/node_modules/event-kit/lib/emitter.js:157
(anonymous) @ /usr/share/atom/resources/app.asar/node_modules/text-buffer/lib/text-buffer.js:1329
Promise resolved (async)
module.exports.TextBuffer.saveTo @ /usr/share/atom/resources/app.asar/node_modules/text-buffer/lib/text-buffer.js:1327
module.exports.TextBuffer.save @ /usr/share/atom/resources/app.asar/node_modules/text-buffer/lib/text-buffer.js:1292
save @ /usr/share/atom/resources/app.asar/src/text-editor.js:1168
\_safeFormatCodeOnSave.takeUntil.toPromise.then @ /home/despairblue/.atom/packages/atom-ide-ui/modules/atom-ide-ui/pkg/atom-ide-code-format/lib/CodeFormatManager.js:174
Promise resolved (async)
editor*.save @ /home/despairblue/.atom/packages/atom-ide-ui/modules/atom-ide-ui/pkg/atom-ide-code-format/lib/CodeFormatManager.js:174
promisify @ /usr/share/atom/resources/app.asar/src/pane.js:887
promisify @ /usr/share/atom/resources/app.asar/src/pane.js:1264
saveItem @ /usr/share/atom/resources/app.asar/src/pane.js:887
autosavePaneItem @ /usr/share/atom/resources/app.asar/node_modules/autosave/lib/autosave.js:55
handleBlur @ /usr/share/atom/resources/app.asar/node_modules/autosave/lib/autosave.js:20
didBlurHiddenInput @ /usr/share/atom/resources/app.asar/src/text-editor-component.js:1510
/home/despairblue/github/prettier-atom/dist/atomInterface/index.js:59 TypeError: Cannot read property 'sync' of undefined
at getPrettierFileInfoForCurrentFilePath (/home/despairblue/github/prettier-atom/dist/helpers/isFileFormattable.js:9)
at /home/despairblue/github/prettier-atom/node_modules/lodash/lodash.min.js:49
at n (/home/despairblue/github/prettier-atom/node_modules/lodash/lodash.min.js:5)
at /home/despairblue/github/prettier-atom/node_modules/lodash/lodash.min.js:51
at u (/home/despairblue/github/prettier-atom/node_modules/lodash/lodash.min.js:5)
at /home/despairblue/github/prettier-atom/node_modules/lodash/lodash.min.js:51
at n (/home/despairblue/github/prettier-atom/node_modules/lodash/lodash.min.js:5)
at /home/despairblue/github/prettier-atom/node_modules/lodash/lodash.min.js:63
at n (/home/despairblue/github/prettier-atom/node_modules/lodash/lodash.min.js:5)
at /home/despairblue/github/prettier-atom/node_modules/lodash/lodash.min.js:51
at u (/home/despairblue/github/prettier-atom/node_modules/lodash/lodash.min.js:5)
at /home/despairblue/github/prettier-atom/node_modules/lodash/lodash.min.js:51
at n (/home/despairblue/github/prettier-atom/node_modules/lodash/lodash.min.js:5)
at /home/despairblue/github/prettier-atom/node_modules/lodash/lodash.min.js:63
at n (/home/despairblue/github/prettier-atom/node_modules/lodash/lodash.min.js:5)
at /home/despairblue/github/prettier-atom/node_modules/lodash/lodash.min.js:97
at n (/home/despairblue/github/prettier-atom/node_modules/lodash/lodash.min.js:5)
at /home/despairblue/github/prettier-atom/node_modules/lodash/lodash.min.js:63
at /home/despairblue/github/prettier-atom/node_modules/lodash/lodash.min.js:49
at attemptWithErrorNotification (/home/despairblue/github/prettier-atom/dist/formatOnSave/index.js:14)
at /home/despairblue/github/prettier-atom/dist/atomInterface/index.js:57
at Generator.next (<anonymous>)
at step (/home/despairblue/github/prettier-atom/node_modules/babel-runtime/helpers/asyncToGenerator.js:17)
at /home/despairblue/github/prettier-atom/node_modules/babel-runtime/helpers/asyncToGenerator.js:35
at new Promise (<anonymous>)
at new F (/home/despairblue/github/prettier-atom/node_modules/core-js/library/modules/\_export.js:36)
at /home/despairblue/github/prettier-atom/node_modules/babel-runtime/helpers/asyncToGenerator.js:14
at attemptWithErrorNotification (/home/despairblue/github/prettier-atom/dist/atomInterface/index.js:65)
at safeFormatOnSaveIfAppropriate (/home/despairblue/github/prettier-atom/dist/formatOnSave/index.js:14)
at /home/despairblue/github/prettier-atom/dist/main.js:38
at Generator.next (<anonymous>)
at step (/home/despairblue/github/prettier-atom/node_modules/babel-runtime/helpers/asyncToGenerator.js:17)
at /home/despairblue/github/prettier-atom/node_modules/babel-runtime/helpers/asyncToGenerator.js:35
at new Promise (<anonymous>)
at new F (/home/despairblue/github/prettier-atom/node_modules/core-js/library/modules/\_export.js:36)
at /home/despairblue/github/prettier-atom/node_modules/babel-runtime/helpers/asyncToGenerator.js:14
at lazyFormatOnSave (/home/despairblue/github/prettier-atom/dist/main.js:42)
at subscriptions.add.editor.getBuffer.onWillSave (/home/despairblue/github/prettier-atom/dist/main.js:111)
at Function.module.exports.Emitter.simpleDispatch (/usr/share/atom/resources/app.asar/node_modules/event-kit/lib/emitter.js:25)
at Emitter.<anonymous> (/usr/share/atom/resources/app.asar/node_modules/event-kit/lib/emitter.js:154)
at Emitter.module.exports.Emitter.emitAsync (/usr/share/atom/resources/app.asar/node_modules/event-kit/lib/emitter.js:157)
at /usr/share/atom/resources/app.asar/node_modules/text-buffer/lib/text-buffer.js:1329

</details>

What I did was:
1. apm develop prettier-atom
2. cd github/prettier-atom
3. git checkout rob/major-simplifications
4. yarn
5. yarn start build

robwise:
@despairblue that's the correct set of steps, yes. I'm not getting that locally. Are you sure your project's version of prettier is up to date?

robwise:
@kachkaev Have you had the same issues as @despairblue?

despairblue:
@robwise Upgrading my local prettier to 1.14 fixed it. Seems 1.8 is missing the `getFileInfo` function.

But it ignores my prettier settings in the package.json.

kachkaev:
I just installed this PR's dev version of `prettier-atom` for a brand-new Atom instance on a MacOS and all worked well for me! I can see `yaml` formatting support that was added in 1.14 and also plugin support (both for globally and locally installed prettier) 🎉 

There is just one minor glitch I've noticed: when a file is being formatted by a plugin (e.g. [`@prettier/plugin-php`](https://github.com/prettier/plugin-php) or [`prettier-plugin-elm`](https://github.com/gicentre/prettier-plugin-elm)) and there is a syntax error, an exception shows up in the UI saying `prettier-atom failed: undefined`.

screen shot 2018-08-02 at 10 58 23
_(`====` does not make sense)_

screen shot 2018-08-02 at 10 37 53
_(two spaces before `output =` are not allowed)_

This does not happen for built-in languages like yaml or javascript. I guess the difference is that the built-in languages throw `SyntaxError` while the plugins I tried throw other types of errors. @robwise do you think that the exception should be suppressed in the UI or should the plugin developers thow a specific type of error?

robwise:
> @robwise do you think that the exception should be suppressed in the UI or should the plugin developers thow a specific type of error?

Hmm, this is a good point. They really should be attempting to throw the proper error, otherwise there's not really anything we can do to be more specific. Even just a generic "<Language> Syntax Error" would be better than what they're doing.

If we suppressed the exception, then it would just look as if nothing was happening and that prettier-atom was broken, so I definitely wouldn't do that as we'll get spammed with issues we can't do anything about.

On the second part, all good suggestions. I will implement.

kachkaev:
> If we suppressed the exception, then it would just look as if nothing was happening 

That was the case for me when it came to built-in languages.

IMO showing a red notification each time a file is saved with a syntax error is a bit too ‘loud’ and may become annoying. However, that may happen rarely in real life, so I’m exaggerating the problem.

Nevertheless what if the SyntaxError notification is yellow or even blue, not red? Prettifying is a side-effect of save, so a red error _on save_ intuitively suggests that the file could not be written until you read what the message says.

robwise:
@kachkaev Okay can you check it out now? I've changed it so that the error only shows up once per start of Atom. I think since it just shows up the once, it's okay to just have the Red version. I admit it was too much when it would come up after every save, but now I think it's okay.

robwise:
@despairblue Does putting your settings in a .prettierrc file instead change anything? The lookup of your settings is now being handled by Prettier, so this should not longer be something that prettier-atom has any control over.

kachkaev:
Thanks for making the changes @robwise.

I tested things again and noticed these things:
* Despite that my test project does not have Prettier so the global instance is used, the error says `Your project's Prettier version is not compatible with prettier-atom. Prettier must be >= 1.13.4` (rather than `Your globally installed ...`). The error is shown only once per session though, which is good.

* There is a small typo worth fixing in message details: `toupdate to` → `to update to`

* When Prettier is up-to-date, editing a malformed `.elm` file still triggers `prettier-atom failed: undefined`, while formatting a malformed `.js` file fails on save silently.

robwise:
> Despite that my test project does not have Prettier so the global instance is used, the error says Your project's Prettier version is not compatible with prettier-atom. Prettier must be >= 1.13.4 (rather than Your globally installed ...). The error is shown only once per session though, which is good.

This is strange as it's not supposed to use your global version. It uses either the local project's version or it uses the bundled version. The bundled version should be up to date unless you're using the dev version of the package and haven't installed the latest.

> There is a small typo in message details: toupdate to → to update to.

Willfix

> When Prettier is up-to-date, editing a malformed .elm file still triggers prettier-atom failed: undefined. Formatting a malformed .js file fails on save silently with no feedback from Prettier.

Will check up on this

robwise:
> When Prettier is up-to-date, editing a malformed .elm file still triggers prettier-atom failed: undefined. Formatting a malformed .js file fails on save silently with no feedback from Prettier.

@kachkaev I can't reproduce this one. If it's a normal syntax error from Prettier (as opposed to one of these weird "undefined" errors it throws which causes a full on popup), we display it via the `linter` package (or nuclide if you're using that instead). Maybe you aren't noticing the error in the linter? It should show up as a red underline.

robwise:
Okay I changed the message to not mention the local package specifically since this isn't always where it's pulling from, now it's just generic.

kachkaev:
I see a change in the error message, thank you. Although it can be still improved by only mentioning the local prettier _or_ the global one, I think it's good enough already!

It's strange that you can't reproduce `prettier-atom failed: undefined` error message as I'm getting it all the time. Here is how you may get to the error:

1. 
    ```bash
    npm i -g prettier prettier-plugin-elm
    mkdir test-project
    cd test-project

    cat <<EOF > broken.js
    output "hello world"
    EOF

    cat <<EOF > working.js
    output = "hello world"
    EOF


    cat <<EOF > Broken.elm
    output "hello world"
    EOF

    cat <<EOF > Working.elm
    output = "hello world"
    EOF

    atom -d .
    ```

2.  Cmd+save in all files
     * `broken.js` - nothing changes (built-in linter highlights the row)
     * `working.js` - semicolon added
     * `Broken.elm` - error shows up: `prettier-atom failed: undefined`
     * `Working.elm` - morphs to:

        ```elm
        module Main exposing (..)
    
    
        output =
            "hello world"
        ```

The behaviour does not depend on the presense of Elm plugin in Atom. The same happens with `@prettier/plugin-php` and `*.php` files.

robwise:
> It's strange that you can't reproduce prettier-atom failed: undefined error message 

Oh I was saying I can't reproduce where you don't get any feedback when formatting a bad JS file as I do still get a linter error. I think with the other languages like elm or php, the plugins or whatever is parsing is not throwing the error that Prettier normally expects so it just calls it undefined?

robwise:
>  So unless this is the case, the reason for this undefined popup can be in that plugins throw Error instead of SyntaxError.

Yes, sorry, this is the same thing as what I was attempting to say. I think we're on the same page here.

> Can you suppress that popup someone (at least for the time being)? I a real scenario, users of prettier plugins will also have linters installed for those extra languages. So when no formatting on save happens, something will be highlighted in Elm and PHP by that other plugin.

Hmm, perhaps we can have a special option as sort of a band-aid until plugins are better supporting errors? Some ideas:

- "Don't show errors when formatting": probably too extreme
- "Don't show `undefined` errors when formatting": maybe this is what we want, although it might be a bit hard to understand if you aren't familiar with the issue at hand.
- "Suppress errors for scopes": then you provide a list of Atom scopes where you want to suppress errors
- "Suppress `undefined` errors for scopes": same as above but only suppresses the `undefined` error

kachkaev:
I agree with you that some options look a bit hard to explain. What if we simply suppress undefined errors without any options and console.log them in Atom’s developer tools? This will make prettier’s behavior look the same for any language at start and we can think of better UX after getting feedback from people.

I don’t exclude that this false positive error message originates from the plugins and has to be fixed upstream. However, given that it may take too much time and the current behavior may feel frustrating, it feels like implementing a special case for undefined is the least evil.

No matter how hard we try to cover edge cases and make the plugin ideal, I’m sure there will be scenarios that we’ve missed. The redesign you were working on is already a breakthrough and I feel really keen to see it on the battlefield! The faster this happens, the quicker the community will polish things up. Iteration over perfection ;–)

robwise:
Yes, I agree on all points! I will implement this (suppress `undefined` errors and console.log them) and then I think we're ready for release!

robwise:
Okay all set, ready to ship finally?

kachkaev:
party parrot party parrot party parrot party parrot party parrot party parrot

billyjanitsch:
Hi! I authored one of the issues that this PR is meant to close (#401).

I understand why you'd want to simplify this plugin to reduce the maintenance burden, but FWIW, I'm a bit sad to see this merged. As I mentioned in https://github.com/prettier/prettier-atom/issues/401#issuecomment-376993090, my primary use case for this plugin is to format one-off files that are not part of a project and therefore don't have a Prettier config (e.g. individual markdown documents, or small gists). If I understand this PR correctly, this is no longer possible?

kachkaev:
@billyjanitsch this change should allow your use case. Just as before, `prettier-atom` tries a project-scoped instance of prettier, then falls back to the global prettier and finally uses a built-in prettier if none exist.

If you want some config to be applied outside your project, simply create `.prettierconfig` in a place like your home folder or even disk root dir. The plugin is no longer responsible for spoon-feeding options into prettier, but it does not mean you cannot configure it!

robwise:
Please let me know if the flow @kachkaev described does not work for you, as I would consider that a bug and will fix ASAP.

PR #425, opened 2018-06-12
Update status-color so it respects theme colors for text-success

chrisdrackett:
Right now if a theme changes @text-color-success this package will not use the value from the theme. This update makes sure that the color set in the theme will be used.


robwise:
@chrisdrackett Wasn't this already addressed in https://github.com/prettier/prettier-atom/pull/400?

chrisdrackett:
@robwise in my testing (and I'm no expert, so maybe I'm missing something) in my theme if I set `@text-color-success` to another value it will not get picked up by prettier and it will still use the default color. However if I use this branch my theme is able to adjust the text color. I'm guessing that the prettier code is evaluated before my theme css can come in and overwrite?

styles/statusTile.less
```diff
@@ -12,7 +12,7 @@ div {
 
   // If Format on Save is enabled, make the label more visible.
   &[data-prettier-format-on-save='enabled'] {
-    color: @text-color-success;
+    // color is handled by setting `text-success`
```
robwise: Let's change this to say "// color is handled by adding or removing the `text-success` class when updating the status tile"

chrisdrackett:
@robwise things should be good now!

robwise:
@chrisdrackett Can you make the commit messages conform to the conventional changelog angular standard? Otherwise they won't be picked up in the changelog

chrisdrackett:
@robwise I'm happy to help out but I'm not sure what that is and don't have a lot of free time right now to figure it out

chrisdrackett:
@robwise ok, took a crack, hopefully this works

robwise:
@chrisdrackett Fixed for you in #435, still had some old commits in this PR that were not properly named

PR #454, opened 2018-09-15
feat(Add tslint to settings and execution): Add support for prettier-tslint

### LINKED ISSUE #302: Support prettier-tslint
I would like to request support for [`prettier-tslint`](https://github.com/azz/prettier-tslint).

jeffrafter:
Currently, tslint support cannot be mixed with prettier directly. This adds support using `prettier-tslint` by following the patterns of the embedded prettier-eslint support.

Fixes #302


jeffrafter:
Thanks @robwise I think you are right... I'm open to making it more _either-or_ and then documenting that. Really whatever you recommend 💃 

I also wanted to note that since I have started using my fork I've noticed the TypeScript server crashing occasionally. I don't know that it is related, but I want to be cautious about it.

robwise:
Sorry for the slow reply, have been traveling.

> Thanks @robwise I think you are right... I'm open to making it more either-or and then documenting that. Really whatever you recommend 💃

I've been trying to think of a good solution, I came up with two alternatives:

- a dropdown Atom option that lets you select `Prettier, prettier-eslint, prettier-tslint`
- treat it like we do `prettier-stylelint` where we just hardcode certain scopes to automatically use `prettier-tslint` if we detect we're in a TS lint file and the user has enabled it.

Do you think the second one would work? That would be more consistent because we already do it like that for prettier-stylelint.

> I also wanted to note that since I have started using my fork I've noticed the TypeScript server crashing occasionally. I don't know that it is related, but I want to be cautious about it.

Hmm, I wonder if it's making calls too frequently to the TS server or something? Do you have it set to constantly lint, or wait until you try and save the file?

robwise:
There's still the case where you don't have either local or global install and it uses a version that comes bundled with prettier-atom itself. That's a lot of falling back!

I think the way it works now is that we basically don't bother detecting whether you have prettier-eslint locally/globally or not when determining whether or not to use it, it is determined entirely by the prettier-atom config option ([`shouldUseEslint`](https://github.com/prettier/prettier-atom/blob/master/src/executePrettier/executePrettierOnBufferRange.js#L69)). 

In fact, we only start looking for where it's installed when its vanilla Prettier ([`getPrettierInstance`](https://github.com/prettier/prettier-atom/blob/master/src/executePrettier/executePrettierOnBufferRange.js#L23)). If it's prettier-eslint, we just use [the version that we bundle with prettier-atom](https://github.com/prettier/prettier-atom/blob/master/src/executePrettier/executePrettierOnBufferRange.js#L3).

So, I guess, I think what's easiest is if we replace the `ESLint Integration` config option with some type of `Formatter` dropdown with three values (`Prettier, prettier-eslint, prettier-tslint`) and determine which one to use based off of this value.

jeffrafter:
I'll give that a shot 🚀

Crimees:
Happy new year everyone. What's the latest on this? It would save me from moving over to VSCode !

robwise:
Given tslint's integration into ESLint and our discussion in https://github.com/prettier/prettier-atom/issues/302, I'm going to close.

Issue #330, opened 2017-12-28
Slow Activation Time


haroldtreen:
## Current Behavior

`prettier-atom` is showing up as my slowest package, with an activation time nearly 2x as long as the subsequent package.

![image](https://user-images.githubusercontent.com/1745854/34421969-95732d3c-ebe0-11e7-8f14-9627d06aa28f.png)

## Expected Behavior

I would expect `prettier-atom` to boot closer to 100 ms? It seems to be an outlier, and I would like to understand why.

## Debug Info

```
Atom version: 1.23.1
prettier-atom version: 0.43.1
prettier version: 1.9.2
prettier-eslint version: 8.6.0
prettier-atom configuration: {
  "useEslint": false,
  "useStylelint": false,
  "useEditorConfig": true,
  "formatOnSaveOptions": {
    "enabled": false,
    "respectEslintignore": true,
    "showInStatusBar": false,
    "javascriptScopes": [
      "source.js",
      "source.jsx",
      "source.js.jsx",
      "source.babel",
      "source.js-semantic",
      "text.html.basic",
      "text.html.vue"
    ],
    "typescriptScopes": [
      "source.ts",
      "source.tsx",
      "source.ts.tsx"
    ],
    "cssScopes": [
      "source.css",
      "source.less",
      "source.css.less",
      "source.scss",
      "source.css.scss",
      "source.css.postcss"
    ],
    "jsonScopes": [
      "source.json"
    ],
    "graphQlScopes": [
      "source.graphql"
    ],
    "markdownScopes": [
      "source.md",
      "source.gfm",
      "text.md"
    ],
    "excludedGlobs": [],
    "whitelistedGlobs": [],
    "isDisabledIfNotInPackageJson": false,
    "isDisabledIfNoConfigFile": false
  },
  "prettierOptions": {
    "singleQuote": false,
    "bracketSpacing": true,
    "semi": true,
    "useTabs": false,
    "jsxBracketSameLine": false,
    "printWidth": 80,
    "tabWidth": "auto",
    "trailingComma": "none",
    "parser": "babylon"
  },
  "prettierEslintOptions": {
    "prettierLast": false
  }
}
```

OS: MacOS 10.13.2

I also experimented with 

`atom --profile-startup .`

I've attached that profile:

[CPU-20171228T150504.cpuprofile.zip](https://github.com/prettier/prettier-atom/files/1592010/CPU-20171228T150504.cpuprofile.zip)

robwise:
[...]

PR #485, opened 2019-02-14
defer activating the package until the window is loaded.

akonwi:
Using the latest version of atom (1.34.0 at this time), this package is consistently the slowest to activate. I've recorded the average activation time for this package in two scenarios;

1. A window with one project open and reloaded 4 times: average activation time is 321ms
2. A window with no project open and reloaded 4 times: average activation time is 222ms

Using this activation hook, atom will defer activating this package until the window is ready to use and then there's no impact to startup time. Since this package doesn't need to do heavy setup, I think it should wait as long as possible to activate anyway.

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,6 +17,9 @@
     "url": "https://github.com/prettier/prettier-atom/issues"
   },
   "license": "MIT",
+  "activationHooks": [
+    "core:loaded-shell-environment"
+  ],
   "dependencies": {
     "atom-linter": "^10.0.0",
     "atom-package-deps": "^4.6.2",


akonwi:
Woops. Sorry, I skimmed the contributing guide and missed that part. I've updated the commit

robwise:
No worries, thanks so much for the PR :)

akonwi:
I just noticed #330. I guess this resolves that as well. After reading the suggestions there and looking at the code, I'm pretty confident that all the `lazy*` functions can be removed because they don't gain much when the package is activated outside of the initial window startup, which is so messy and suboptimal.

PR #487, opened 2019-02-15
Cleanup of lazy loading code

akonwi:
After further profiling of the `activate` function, I realized it still took ~300ms, which is alright when it's deferred. After removing the lazy loading wrappers and combining those `subscriptions.add` calls into one, I shaved off >100ms, which is further improvement.

`loadPackageDeps` is the bottleneck now. It's accounting for over 100ms while the rest of the activation is about 15ms.

Is the linter package a hard dependency? I'd suggest deferring the check for it until after activation or during the `initialize` phase of the package because the service consumption happens whenever the providing package is activated. So prettier-atom can activate independently, and a user can install linter much later on and automatically see the messages from this package.

### DIFF
--- a/dist/main.js
+++ b/dist/main.js
@@ -1,73 +1,24 @@
 'use strict';
 
-var _asyncToGenerator2 = require('babel-runtime/helpers/asyncToGenerator');
-
-var _asyncToGenerator3 = _interopRequireDefault(_asyncToGenerator2);
-
-function _interopRequireDefault(obj) { return obj && obj.__esModule ? obj : { default: obj }; }
-
 // eslint-disable-next-line import/no-extraneous-dependencies, import/no-unresolved
 const { CompositeDisposable } = require('atom');
 const config = require('./config-schema.json');
 const { createStatusTile, updateStatusTile, updateStatusTileScope, disposeTooltip } = require('./statusTile');
 const linterInterface = require('./linterInterface');
+const format = require('./manualFormat'); // eslint-disable-line global-require
+const formatOnSave = require('./formatOnSave'); // eslint-disable-line global-require
+// eslint-disable-next-line global-require
+const warnAboutLinterEslintFixOnSave = require('./warnAboutLinterEslintFixOnSave');
+// eslint-disable-next-line global-require
+const displayDebugInfo = require('./displayDebugInfo');
+// eslint-disable-next-line global-require,prefer-destructuring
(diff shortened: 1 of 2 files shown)


robwise:
Besides the weird diff noise, this looks good to me, it's logical that we can now get rid of the lazy-loading code since the whole package essentially is lazy loaded now.

The package doesn't work without the `linter` package, no, but it doesn't theoretically need it until it has to format something. However, I believe the way all of that dependency checking stuff is done is via this automatic thing: https://github.com/prettier/prettier-atom/blob/master/package.json#L92-L94 (enabled by https://www.npmjs.com/package/atom-package-deps), so the implementation is buried in that package.

Do you know why that diff noise is happening?

akonwi:
I cleaned up the diff. I guess I did something weird while updating my fork. The generated travis build failed oddly but a new one was kicked off that passed.

> The package doesn't work without the `linter` package, no, but it doesn't theoretically need it until it has to format something

I'm proposing that if indeed we want to ensure that Linter is available for communicating feedback to users, then hold activation until `require('atom-package-deps').install('prettier-atom')` completes. I have made some more changes do that and can open another pr if you are open to that idea.

robwise:
> I'm proposing that if indeed we want to ensure that Linter is available for communicating feedback to users, then hold activation until `require('atom-package-deps').install('prettier-atom')` completes. I have made some more changes do that and can open another pr if you are open to that idea.

Yeah I'm open to that, sure. Okay LGTM

akonwi:
@robwise So it turns out the atom-package-deps' `install` function always resolves the promise, even if the user declines to install. Because it's not actually enforcing the dependency, maybe there should be a backup plan for communicating feedback. Errors are the only messages shown in linter right? Perhaps in addition to `console.error`, we show an error notification with the prettier message.

robwise:
There are certain types of errors that Prettier can throw in strange cases (usually a bug) that do not have the normal syntax error look to the error object, and then we detect that and put it inside of a normal Atom error notification dialog (since linter requires line positions and such).

But yeah besides that, we are just showing in linter. We used to not use linter and use just the error notification for syntax errors but we got a lot of complaints about that being too intrusive/annoying. We really just require you to use the linter package for this since it's such a well-established pattern for this type of a plugin.

robwise:
I just published this one like a minute ago btw, in case you were testing it with the non-dev version.

akonwi:
Gotcha. Yea that makes sense. I'll leave it as it is then

PR #498, opened 2019-05-14
fix(prettierignore): Pass relative instead of absolute path to prettier.getFileInfo

chancedickson:
Currently there is a bug in how prettier-atom determines if a file is ignored or not. When calling prettier.getFileInfo, it's being passed the absolute path of the current file instead of a relative path. This causes bugs with negations in .prettierignore. An example of the bug is at https://github.com/chancedickson/prettier-atom-bug-example. The only file that should be formatted is foo/bar/should-format.js (which can be confirmed by running `prettier -c **/*` in the root of the repo), however should-not-format.js and foo/should-not-format.js will be formatted-on-save by prettier-atom as well. This commit fixes that issue.


chancedickson:
@robwise Think I have it fixed up. Let me know what you think 😁

src/helpers/isFileFormattable.js
```diff
@@ -8,13 +9,26 @@ const { findCachedFromFilePath } = require('./general');
 const getNearestPrettierignorePath = (filePath: FilePath): ?FilePath =>
   findCachedFromFilePath(filePath, '.prettierignore');
 
+const getNearestPrettierignoreParentDirPathRelativeToCurrentFilePath = (editor: TextEditor): string => {
```
robwise: Sorry this was my fault again, I got the name like backwards, it should be: `getCurrentFilePathRelativeToNearestPrettierignoreParentDir`

chancedickson:
@robwise Ah, didn't even think about that. Fixed!

johnsenpeder:
Ignores doesn't seem to work at all for me anymore now with the latest version? :thinking: 
(They worked in the previous version :man_shrugging:)

chancedickson:
@johnsenpeder Investigating. Could you share your .prettierignore file?

johnsenpeder:
@chancedickson it was originally:
```
client/app/less
client/app/src
client/app/vendor
```
but also tried:
```
/client/app/less
/client/app/src
/client/app/vendor
```

PR #509, opened 2019-07-19
fix(debug): read-pkg-up@6 renamed pkg to package

josmardias:
BC break on read-pkg-up@6, renamed { pkg: ...} to { package: ...}
https://github.com/sindresorhus/read-pkg-up/releases/tag/v6.0.0

fixes #508

PR #511, opened 2019-07-25
fix(deps): correctly use new pkg-up, fixup f5a0b18, #507

### LINKED ISSUE #510: Only format if Prettier in dependencies does not work
The option “Only format if Prettier is found in your project's depedencies” does not detect Prettier correctly and the formatter does not run.

image

It works when downgrading to 0.56.6, but does not work anymore on 0.57.0.

The regression is caused by an incorrect dependency bump in f5a0b18 #51 

read-pkg-up [in the new version](https://github.com/sindresorhus/read-pkg-up/releases) renames a key from `pkg` to `package`:

image

Prettier is very definitely in project’s dependencies, it was there for many months, and the option worked before. It does not work with all my projects now. (`"prettier": "^1.18.2"`)

This has began since a month or two ago, until then the addon worked great. Both functionality-wise and timeline-wise this seems quite similar to #502.

Atom’s console has no relevant messages neither on start-up, nor at the time when I save. Manually triggering Prettier: Format does format the file. Unchecking the option leads to the plugin formatting on save.

This is not related to symlinks or directory permissions, a clean project clone on the desktop produces the same results.

<details>

Atom version: 1.39.0
prettier-atom version: 0.57.1
prettier: bundled
prettier version: 1.18.2
prettier-eslint version: 9.0.0
prettier-atom configuration:
```json
{
  "formatOnSaveOptions": {
    "enabled": true,
    "jsonScopes": [
      "-"
    ],
    "showInStatusBar": true,
    "whitelistedGlobs": [],
    "isDisabledIfNotInPackageJson": true,
    "respectEslintignore": true,
    "excludedGlobs": [],
    "isDisabledIfNoConfigFile": false,
    "ignoreNodeModules": true
  },
  "useEslint": false,
  "useStylelint": false,
  "prettierEslintOptions": {
    "prettierLast": false
  }
}
```

Search terms: isDisabledIfNotInPackageJson #43

</details>

denis-sokolov:
Closes #510.

### DIFF
--- a/dist/formatOnSave/isPrettierInPackageJson.js
+++ b/dist/formatOnSave/isPrettierInPackageJson.js
@@ -12,9 +12,9 @@ const {
   shouldUseEslint
 } = require('../atomInterface');
 
-const hasPackageDependency = packageName => _.flow(_.get('pkg.dependencies'), _.has(packageName));
+const hasPackageDependency = packageName => _.flow(_.get('package.dependencies'), _.has(packageName));
 
-const hasPackageDevDependency = packageName => _.flow(_.get('pkg.devDependencies'), _.has(packageName));
+const hasPackageDevDependency = packageName => _.flow(_.get('package.devDependencies'), _.has(packageName));
 
 const hasPackage = packageName => _.overSome([hasPackageDependency(packageName), hasPackageDevDependency(packageName)]);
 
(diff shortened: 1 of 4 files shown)


josmardias:
I was looking for this issue when I ended up creating #509.
But I was not aware that the formatting stopped only when that option was "on".
Thansk for that, good job!

PR #516, opened 2019-10-04
fix(global-prettier-resolution): add fallback for NVM users

### LINKED ISSUE #448: Global Version of Prettier Not Resolved in Linux
If a local version of Prettier is not found, prettier-atom is supposed to fall back to a global version if one is available. This does not work on Linux.

It's not showstopping because we can then further fall back to a bundled version that comes with prettier-atom, but it should be fixed.

Reported by @krugar

kachkaev:
Closes #448 

When Prettier is not found in the current project, the extension first attempts to find a global instance of it and finally falls back to the built-in instance. The second step is important for those who want to use Prettier with plugins, but does not want to see local `node_modules` in each project just because of it. Searching for global Prettier was added in https://github.com/prettier/prettier-atom/pull/397 by me; we use this feature in [litvis](https://github.com/gicentre/litvis#automatic-code-formatting).

Recently, my colleague @jwoLondon discovered that global Prettier does not get resolved when Node and NPM are installed via [NVM](https://github.com/nvm-sh/nvm). This setup method is pretty prevalent and is especially helpful when a user does not have admin rights (e.g. on managed company laptops and in education environments such as university labs).

We investigated the problem and discovered that it comes from [global-modules](https://www.npmjs.com/package/global-prefix), which does not work as reliably as the `npm get prefix` command. Related issues:

- https://github.com/jonschlinkert/global-prefix/issues/19
- https://github.com/jonschlinkert/global-prefix/issues/20
- https://github.com/jonschlinkert/global-prefix/issues/21

This PR introduces a fallback search mechanism for global node modules, which should at least help NVM users, but maybe solves the issue in other edge cases too. I tested it by creating a new macOS user, installing NVM and then running `npm install --global prettier prettier-plugin-elm`. Debug info no longer shows `prettier: bundled` – it is now `/Users/tester/.nvm/versions/node/v12.11.1/lib/node_modules/prettier/index.js`. Global plugins work! 🎉

I decided to add the fallback search method instead of replacing the existing one to make sure the fix is non-breaking. Hope it can land the next patch release soon and it is much needed at City, University of London this term!

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,6 +26,7 @@
     "atom-package-deps": "^5.1.0",
     "atom-text-buffer-point": "^1.0.0",
     "atom-text-buffer-range": "^1.0.0",
+    "execa": "^1.0.0",
     "global-modules": "^2.0.0",
     "ignore": "^5.1.2",
     "lodash": "^4.17.13",
--- a/src/helpers/getPrettierPath.js
+++ b/src/helpers/getPrettierPath.js
@@ -1,14 +1,38 @@
 // @flow
 const path = require('path');
+const { sync: execaSync } = require('execa');
 const { findCached } = require('atom-linter');
+const { memoize } = require('lodash');
 const globalModules = require('global-modules');
 const yarnGlobalModules = require('yarn-global-modules')();
 const { findCachedFromFilePath } = require('./general');
 
 const PRETTIER_INDEX_PATH = path.join('node_modules', 'prettier', 'index.js');
 
+// global-prefix package (which is used by global-modules) fails
+// to locate the directory if Node and NPM are installed via NVM
+// https://github.com/jonschlinkert/global-prefix/issues/19
+// The function below applies a fallback search mechanism,
+// which relies on calling 'npm get prefix'
+const getFallbackGlobalModulesPath = memoize(() => {
+  try {
+    const prefix = execaSync('npm', ['get', 'prefix']).stdout;
+
+    // Inspired by global-modules
+    // https://github.com/jonschlinkert/global-modules/blob/7455e827f0ce0366d224aea035a8fecd233aa24a/index.js
+    if (process.platform === 'win32' || process.env.OSTYPE === 'msys' || process.env.OSTYPE === 'cygwin') {
+      return path.resolve(prefix, 'node_modules');
+    }
+    return path.resolve(prefix, 'lib/node_modules');
+  } catch (e) {
+    return '';
+  }
+});
+
 const getGlobalPrettierPath = (): ?FilePath =>
-  findCached(globalModules, PRETTIER_INDEX_PATH) || findCached(yarnGlobalModules, PRETTIER_INDEX_PATH);
+  findCached(globalModules, PRETTIER_INDEX_PATH) ||
+  findCached(yarnGlobalModules, PRETTIER_INDEX_PATH) ||
+  findCached(getFallbackGlobalModulesPath(), PRETTIER_INDEX_PATH);
 
 const getLocalPrettierPath = (filePath: ?FilePath): ?FilePath =>
   findCachedFromFilePath(filePath, PRETTIER_INDEX_PATH);
--- a/dist/helpers/getPrettierPath.js
+++ b/dist/helpers/getPrettierPath.js
@@ -3,8 +3,16 @@
 const path = require('path');
 
 const {
+  sync: execaSync
+} = require('execa');
+
+const {
   findCached
 } = require('atom-linter');
+
+const {
+  memoize
+} = require('lodash');
 
 const globalModules = require('global-modules');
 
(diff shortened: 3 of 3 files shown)


package.json
```diff
@@ -26,6 +26,7 @@
     "atom-package-deps": "^5.1.0",
     "atom-text-buffer-point": "^1.0.0",
     "atom-text-buffer-range": "^1.0.0",
+    "execa": "^1.0.0",
```
kachkaev: The [latest version of `execa`](https://www.npmjs.com/package/execa?activeTab=versions) is 2.0.5, but I'm using 1.0.0 to avoid extra dependencies in yarn.lock. This version of execa has already been there because it's a sub-dependency.

src/helpers/getPrettierPath.js
```diff
@@ -1,14 +1,38 @@
...
 
 const PRETTIER_INDEX_PATH = path.join('node_modules', 'prettier', 'index.js');
 
+// global-prefix package (which is used by global-modules) fails
+// to locate the directory if Node and NPM are installed via NVM
+// https://github.com/jonschlinkert/global-prefix/issues/19
+// The function below applies a fallback search mechanism,
+// which relies on calling 'npm get prefix'
+const getFallbackGlobalModulesPath = memoize(() => {
+  try {
```
robwise: Which line specifically is it that can throw an exception and why? From looking at `execa`, it looks like it can throw errors if you call a bad command, for example. However, we probably don't want to swallow errors in that case like this? 

Do we expect that sometimes the user won't have `npm` on the path or something? If so, we should do a pre-emptive guard for that by executing `which`?

For example, here is roughly the same approach but no try/catch so that we don't swallow unexpected errors:

```js
const npmIsAvailable = () => !_.isEmpty(execaSync("which", ["npm"]).stdout);

const npmPrefix = () => execaSync("npm", ["get", "prefix"]).stdout;

const resolveGlobalNodeModulesFromPrefix = prefix =>
  process.platform === "win32" ||
  process.env.OSTYPE === "msys" ||
  process.env.OSTYPE === "cygwin"
    ? path.resolve(prefix, "node_modules")
    : path.resolve(prefix, "lib/node_modules");

const resolveGlobalNodeModulesWithNpm = () => _.flow(npmPrefix, resolveGlobalNodeModulesFromPrefix);

const getFallbackGlobalModulesPath = _.memoize(() => _.cond([[npmIsAvailable, resolveGlobalNodeModulesWithNpm]]));
```

src/helpers/getPrettierPath.js
```diff
@@ -1,14 +1,38 @@
 // @flow
 const path = require('path');
+const { sync: execaSync } = require('execa');
 const { findCached } = require('atom-linter');
+const { memoize } = require('lodash');
```
robwise: Can we use `lodash/fp` here? Shouldn't actually affect your code but since we're using it everywhere else, it might be confusing if someone else comes along and tries to add some additional lodash imports for something they're doing.

kachkaev: Sure thing, I'll update the PR with `lodash/fp`. I did not notice this was the main flavour of the lib used in the repo.

src/helpers/getPrettierPath.js
kachkaev: > Which line specifically is it that can throw an exception and why?

I'm generally trying to guard the execa call just in case, since the signature of that function does allow for an exception to be thrown. I can't think of a concrete example, but it still feels safer with `try` / `catch`. I would prefer this solution over calling `which` because `which` may (?) not be available on Windows and also because it's another sync CLI call that takes time.

Your solution implies calling `execaSync` every time, even when fallback is not needed. This is what I was tryign to avoid by implementing `getFallbackGlobalModulesPath` the way I did. I'm totally happy to copy-paste your code though because I don't think that an extra CLI call is that expensive. WDYT?

src/helpers/getPrettierPath.js
```diff
@@ -1,14 +1,38 @@
...
+    const prefix = execaSync('npm', ['get', 'prefix']).stdout;
+
+    // Inspired by global-modules
+    // https://github.com/jonschlinkert/global-modules/blob/7455e827f0ce0366d224aea035a8fecd233aa24a/index.js
+    if (process.platform === 'win32' || process.env.OSTYPE === 'msys' || process.env.OSTYPE === 'cygwin') {
+      return path.resolve(prefix, 'node_modules');
+    }
+    return path.resolve(prefix, 'lib/node_modules');
+  } catch (e) {
+    return '';
```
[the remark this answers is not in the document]
kachkaev: You are right, the signature of `getGlobalPrettierPath` can return `null`. However, `findCached` from [`atom-linter`](https://www.npmjs.com/package/atom-linter) requires the first argument (`directory`) to be a string: 

```flow
export function findCached(directory: String, names: String | Array<string>): ?String
```

I still think we should be returning `''` from `resolveGlobalNodeModulesFromPrefix` when prefix is not found, otherwise, `findCached(null, PRETTIER_INDEX_PATH)` can potentially throw (if not now, then in future version).

kachkaev:
Thank you for your feedback @robwise! I'll be happy to do a round of changes to this PR after hearing back from you 😉

src/helpers/getPrettierPath.js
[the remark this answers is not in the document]
robwise: Ah right okay good point, I was wrong

src/helpers/getPrettierPath.js
robwise: Well, the whole thing does get memoized at the end in the code sample I put so there should only be those two calls to execa and then that's it.

But, the fact that `which` may not be present in Windows is concerning. I wasn't thinking about that. Let's merge for now and then we can always address later if we find this becomes a problem (which it probably won't).

robwise:
Oh crap we changed the PR title not the commit, so it's not going to show up in the changelog. But I will cut a release 0.57.3 now.

kachkaev:
Many thanks for cutting the release so quickly @robwise! All works 🎉

No worries at all about the changelog – sorry for not fully understanding what you meant. I thought you'd squash and merge, so the commit message would contain the title of the PR.

robwise:
Oh yeah I'm in the habit of just doing the regular merge commits (personal preference, I like seeing the PRs in the git history). I'm glad this is working for the rest of your team now!

Issue #514, opened 2019-09-03
Editor config


amk221:
I have Format on Save enabled.

When I save, lines are printed at 80 chars.

But when I run prettier on the command line, they are printed at 100 chars.

This is because prettier is respecting the value in my editorconfig.

...but prettier-atom isn't?

<details>
  <summary>debug</summary>
Atom version: 1.40.1
prettier-atom version: 0.57.2
prettier: bundled
prettier version: 1.18.2
prettier-eslint version: 9.0.0
prettier-atom configuration: {
  "formatOnSaveOptions": {
    "enabled": true,
    "isDisabledIfNoConfigFile": true,
    "isDisabledIfNotInPackageJson": true,
    "showInStatusBar": true,
    "respectEslintignore": true,
    "excludedGlobs": [],
    "whitelistedGlobs": [],
    "ignoreNodeModules": true
  },
  "prettierEslintOptions": {
    "prettierLast": false
  },
  "prettierOptions": {
    "singleQuote": true
  },
  "useStylelint": true,
  "useEslint": false
}
</details>

robwise:
prettier-atom is ripping the desired config settings from the nearest `.prettierrc` from the file in question using `resolveConfig`. It looks like at some point they added the ability to pull in editorconfig settings as well but we have to explicitly tell it to do that:

> If options.editorconfig is true and an .editorconfig file is in your project, Prettier will parse it and convert its properties to the corresponding prettier configuration. This configuration will be overridden by .prettierrc, etc. Currently, the following EditorConfig properties are supported:
>
> end_of_line
> indent_style
> indent_size/tab_width
> max_line_length

https://prettier.io/docs/en/api.html#prettierresolveconfigfilepath-options


Therefore I think we can add it maybe as a prettier-atom option where you can opt in or out of passing `editorconfig: true` and have prettier respect your options?

amk221:
It's a bit odd that, the default behaviour when running prettier on the command line is to refer to editor config, and you have to specifically opt out, by using `--no-editorconfig`. And your suggestion is I would have to opt in. Would it be better to just do what prettier does by default?

robwise:
Oh hmm that's a good point. I guess the problem lies in that prettier's `resolveConfig` is acting differently (defaults-wise) than the CLI tool. 

But we can "fix" that problem by automatically opting the user in and then allowing them to explicitly opt out via a prettier-atom setting?

amk221:
[...]

PR #518, opened 2019-10-18
feat: Add option to use configuration from `.editorconfig`

amk221:
Add an option to use configuration from `.editorconfig`, enabled by default.

BREAKING CHANGE: If the user has a `.editorconfig` file, with `max_line_length` then that value will
be used instead of Prettier's `printWidth`. This is the default behaviour of prettier when run on
the command line, but was not the default behaviour of prettier-atom (but it should be, and now is)

#514

### DIFF
--- a/dist/atomInterface/index.js
+++ b/dist/atomInterface/index.js
@@ -26,6 +26,8 @@ const shouldUseEslint = () => getConfigOption('useEslint');
 
 const shouldUseStylelint = () => getConfigOption('useStylelint');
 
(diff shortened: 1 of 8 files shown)


amk221:
Wow, are you sure that was ok to merge?? I was totally expecting some feedback there 😂

robwise:
Yep, commit message looked good, tests were added, CI passed. We are just adding an option that we simply pass through to Prettier after all :)

PR #566, opened 2019-11-04
fix: search and replace 'prettier-eslint' to custom fork

### LINKED ISSUE #565: Can't update to or install v0.58.1
I saw there was an update available and used the atom ui to update the package but it failed and showed this error:

```
Installing “prettier-atom@0.58.1” failed.Hide output…

npm ERR! Invalid dependency type requested: alias

npm ERR! A complete log of this run can be found in:
npm ERR!     /Users/akonwi/.atom/.apm/_logs/2019-11-04T20_21_05_062Z-debug.log
```

I uninstalled the package completely and attempted to just install the latest version and I get the same result and error.

lewisl9029:
This is an alternative implementation of https://github.com/prettier/prettier-atom/pull/564, which
used yarn alias to avoid making code changes. Unfortunately apm is tightly coupled to npm, which
doesn't support the alias feature.

Seems like a naive search and replace of `'prettier-eslint'` to `'@lewisl9029/prettier-eslint'` just worked according to the tests and local testing. @robwise let me know if you notice anything that shouldn't have been changed.

fix #565


src/formatOnSave/isPrettierInPackageJson.js
```diff
@@ -35,7 +35,7 @@ const isPrettierInPackageJson: (editor: TextEditor) => boolean = _.flow(
 const isPrettierEslintInPackageJson: (editor: TextEditor) => boolean = _.flow(
   readContentsOfNearestPackageJson,
   _.overSome([
-    hasPackage('prettier-eslint'),
+    hasPackage('@lewisl9029/prettier-eslint'),
```
robwise: This is subtle but this code here is trying to determine whether to enable prettier-atom format on save or not based on whether or not the user already has prettier, or in this case, prettier-eslint, in their package-json. Therefore, I think the change we need here is to actually cue on _either_ package being present (in the off change they do have your package installed). Either that, or revert this change?

lewisl9029: Ah great catch. Reverting sounds like the way to go. I don't expect my fork to get installed in projects outside of this one. 😅 

Since this repo itself looks like it might be dogfooding this extension, I'll also add prettier-eslint to devDependencies so we don't need any special logic to handle that.

lewisl9029: Pushed up the changes. Let me know if you see anything else weird.

package.json
```diff
@@ -64,6 +63,7 @@
     "jest": "^24.9.0",
     "nps": "^5.9.8",
     "opt-cli": "^1.6.0",
+    "prettier-eslint": "^9.0.0",
```
robwise: @lewisl9029 Is this correct?

lewisl9029: Yep, wanted to make sure this repo itself can dogfood the extension, which requires prettier-eslint to exist in devDependencies.

robwise:
OK I think we're good to go. Can you rebase and force push back up and I'll merge?

lewisl9029:
Updated!

robwise:
Awesome thanks!

